PR #30, opened 2016-12-16
solves the failure of single precision due to twiddle 

tingxingdong:
- add a branch of calling double precision function in pow2.cpp   
- add more test cases in google unit tests 
- change twiddle to float temporarily to solve the unit test failure.


bragadeesh:
thanks @tingxingdong

PR #31, opened 2016-12-16
Fixing cmakefile to honor user provided CMAKE_BUILD_TYPE

kknox:
Summary of proposed changes:
-  Fix build files to properly build both debug and release binaries

The prior behavior always built debug binaries
ignoring visual studio code metadata directories

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,6 @@
 tags
 .tags
 .*.swp
+
+# Visual Studio Code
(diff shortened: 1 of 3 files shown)


kknox:
The failures here are unit tests that should be passing with #30.  The build properly makes release mode binaries.

bragadeesh:
thanks @kknox

PR #33, opened 2017-01-06
add size support > 4096

tingxingdong:
resolves #___

Summary of proposed changes:
-  add large twiddle table (> 4096) 
-  add large size in tester
-  particular size 65536 of double precision fails (single precision passes)


sklam:
With this patch, I am still seeing the following error at size>=8192:

```
/rocFFT/library/src/transform.cpp:55: rocfft_status rocfft_execute(const rocfft_plan, void **, void **, rocfft_execution_info): Assertion `info->workBufferSize >= (execPlan.workBufSize * 2 * plan->base_type_size )' failed.
Aborted (core dumped)
```

See https://gist.github.com/sklam/6ed2a28513587f6c64536652135cd413 for my test script.

PR #39, opened 2017-05-18
generating pow2, 3, 5 kernels 

tingxingdong:
Summary of proposed changes:
-  add a kernel generator in folder /library/src/generator and add it in cmakefile to generate kernels in building rocFFT time
-  change the naming pow2->powX where applicable 
-  add more examples: double precision, large size test (before only 16)
-  add stride support in google unit testers and large pow2 (>4k) device kernels


bragadeesh:
1. Resolve conflicts with my new checkins from this week
2. Make the generator place files in build directory and NOT in source directory
3. Remove rad 7 support in plan.h
4. move GetWGSandNT declaration from twiddles.h and put in another file
5. SB_UNIT calls in pow2_large_entry.h ; change back to 1,1
6. Unsupported length should not cause seg fault
7. plan.cpp change back the large case division logic to original; put the new code in a commented out block
8. Run test suite on AMD & NV for pow2 single precision (All cases); pow3, pow5 single precision (small 1d only) 
9. Run self test on AMD & NV for single precision

tingxingdong:
please ignore the commits above this comments, they are created by rebasing. Nothing new.

tingxingdong:
Until now, rocblas build on both amd and nvidia platform. All single precision passes on amd. while 128, 1024, 4096, 262144 still fails on Nvidia.

bragadeesh:
does these tests pass on NV with the upstream repo?

tingxingdong:
fix the bugs to cause 128, 1024, 4K and 262K failing. It is not a deep bug but a minor one.

tingxingdong:
Now single pow2, pow3, pow5 all pass on Nvidia and AMD platform.,

PR #45, opened 2017-06-14
fix the warning and disable some printing

tingxingdong:
resolves #___

Summary of proposed changes:
-  
-  
-


tingxingdong:
interesting. nvcc does not complain:  "(void)(rw==rw) always true". But hcc does.

bragadeesh:
Hmmm, can you put inside macro
#ifdef __NVCC__
(void)(rw==rw);
#endif

tingxingdong:
then hcc does not see this line at all. However, it rolls back to be  previous case: complain rw is defined but not used.

bragadeesh:
OK, how about

Hmmm, can you put inside macro
#ifdef \_\_NVCC\_\_
(void)(rw==rw);
#else
(void)rw;
#endif

PR #51, opened 2017-08-15
merge oOffset and iOffset calculation into one block

tingxingdong:
resolves #___

Summary of proposed changes:
-  change the kernel generator
-  
-

### DIFF
--- a/library/src/device/generator/generator.kernel.hpp
+++ b/library/src/device/generator/generator.kernel.hpp
@@ -274,14 +274,19 @@ namespace StockhamGenerator
              can be ioOffset, iOffset or oOffset, they are size_t type
           stride_name
              can be stride_in or stride_out, they are vector<size_t> type
+          output
+             if true offset_name2, stride_name2 are enabled
+             else not enabled 
         */
 
-        inline std::string OffsetCalc(const std::string offset_name, const std::string stride_name, bool input = true, bool rc_second_index = false)
+        inline std::string OffsetCalc(const std::string offset_name1, const std::string stride_name1, 
+                                      const std::string offset_name2, const std::string stride_name2,  
+                                      bool output, bool rc_second_index = false)
         {
             std::string str;
 
             /*===========the comments assume a 16-point FFT============================================*/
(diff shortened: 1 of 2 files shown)


library/src/device/generator/generator.kernel.hpp
```diff
@@ -342,21 +347,25 @@ namespace StockhamGenerator
             /*=======================================================*/
             std::string loop;
             loop += "\tif(dim == 1){\n";
-            loop += "\t\t" + offset_name + " += counter_mod*" + stride_name + "[1];\n";
+            loop += "\t\t" + offset_name1 + " += counter_mod*" + stride_name1 + "[1];\n";
+            if (output == true) loop += "\t\t" + offset_name2 + " += counter_mod*" + stride_name2 + "[1];\n";
```
b-sumner: How will this work with the ";" at the end of the preceding line?

tingxingdong: This is the generator, it generates a line of code like this 
`
oOffset += counter_2*stride_out[3] + counter_1*stride_out[2] + counter_mod_1*stride_out[1]; `

b-sumner: I get that, but doesn't line 350 end the generated line with a ";".  And then if output is true, we will have a +... following the ";" in the generated code, which is legal, but doesn't actually do anything.

bragadeesh: @b-sumner I don't see the problem, there's the offset_name2 string before the '+=' , which would look like 'oOffset += ' in the generated line.
@tingxingdong you tested this code right?

tingxingdong: for the generated two lines of code. I do not see the "+" sign.   They should be correct

```
		iOffset += counter_mod*stride_in[1];
		oOffset += counter_mod*stride_out[1];
```

b-sumner: Sorry, you're right.  I misread the code.

PR #58, opened 2017-09-06
generate large 1D pow2, first version

tingxingdong:
-  generate large 1D kernels of pow2 from 8K to 64K (but the CPU calling part need to update, too) 
-  add blockCompute feature in generator
-  modify offsetCalcBlock to make stride as a runtime parameter of kernels
-  remove stride check in generator, as stride is passed as an run-time variable of kernels 
-  this pull request does not change small 1D (<= 4K) kernels.

### DIFF
--- a/library/src/device/generator/generator.main.cpp
+++ b/library/src/device/generator/generator.main.cpp
@@ -23,45 +23,35 @@ using namespace StockhamGenerator;
 
 
 
-rocfft_status initParams (FFTKernelGenKeyParams &params, size_t LEN, size_t STRI)
+rocfft_status initParams (FFTKernelGenKeyParams &params, std::vector<size_t> fft_N,  bool blockCompute, BlockComputeType blockComputeType)
 {
             /* =====================================================================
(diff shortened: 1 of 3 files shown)


library/src/device/generator/generator.main.cpp
```diff
@@ -363,33 +353,102 @@ void WriteCPUFunctionPool(std::vector<size_t> support_list)
...
-
     std::string programCode;
     rocfft_precision pr = rocfft_precision_single;
     FFTKernelGenKeyParams params;
+    BlockComputeType blockComputeType;
 
-    initParams(params, len, stride);
-
-    switch(pr)
+    if(len > 4096) // must decompose into two kernels to do the transform : e.g 8192 = 64*128
```
tingxingdong: 1D threadshold

tingxingdong:
need add device function TW3step in common.h, TW3step is used in large 1D.  Otherwise, "error: use of undeclared identifier 'TW3step'"

PR #71, opened 2017-10-31
Build framework refactor

kknox:
Summary of proposed changes:
-  introducing an easy to use install.sh bash script to build on ubuntu based systems
-  builds on rocm and cuda based targets
-  eliminates most externalproject_add() based code 
    - dependencies are pulled out into a separate cmake file
- upgraded cmake required version to v3.5
- documentation updates


kknox:
This improves the command line build experience.

The PR fails because the Jenkinsfile has not been updated yet.  That is coming next.

bragadeesh:
thanks @kknox

Issue #67, opened 2017-10-26
Problems with double 1D complex interleaved for moderate to large sizes


WendyThrash:
### What should happen
-rider should run correctly for doubles as well as floats.

### What really happens
-It doesn't, for doubles.  Typically gets output[0].real() = =nan, not N.
As far as I can tell, all sizes > 2048 are broken except for powers of 2.
4096 is also broken, but gets 3.88098e-304 rather than -nan.
8192, 16384, etc. are OK.
 

### How to reproduce
-  ./rocfft-rider --double -o  -p 10 -x 2160

### Environment
| Hardware | description |
|-----|-----|
| GPU | device string |
| CPU | device string |
VGA compatible controller: Advanced Micro Devices, Inc. [AMD/ATI] Ellesmere [Radeon RX 470/480/570/580] (rev e7)
vendor_id	: AuthenticAMD
cpu family	: 23
model		: 1
model name	: AMD Ryzen 7 1800X Eight-Core Processor
stepping	: 1
microcode	: 0x8001129

| Software | version |
|-----|-----|
| ROCK | v0.0 |
| ROCR | v0.0 |
| HCC | v0.0 |
 4.11.0-kfd-compute-rocm-rel-1.6-180 #1 SMP Tue Oct 10 08:15:38 CDT 2017 x86_64 x86_64 x86_64 GNU/Linux
HCC clang version 6.0.0  (based on HCC 1.0.17412-f590a25-821e6d8-64e7fc7 )
| Library | v0.0 |

tingxingdong:
Passed on my side

tim@tim-hsa:~/FFT/rocFFT_BUILD/clients-build/rider-build/staging
$ ./rocfft-rider --double -o -p 10 -x 2160

precision: double
transform type: complex forward
result placement: not in-place

input array type: complex interleaved
output array type: complex interleaved

dimensions: 1
lengths: 2160
batch size: 1

input offset: 0
output offset: 0

input strides: 1
output strides: 1
input distance: 2160
output distance: 2160

scale: 1

number of blocks = 2, number of threads = 126
number of blocks = 2, number of threads = 128
number of blocks = 2, number of threads = 126
number of blocks = 2, number of threads = 128
number of blocks = 2, number of threads = 126
number of blocks = 2, number of threads = 128
number of blocks = 2, number of threads = 126
number of blocks = 2, number of threads = 128
number of blocks = 2, number of threads = 126
number of blocks = 2, number of threads = 128
[...]

PR #72, opened 2017-11-01
fix bugs for double precsion 

tingxingdong:
resolves issue #67 
Summary of proposed changes:
-  now double pass for large1D of pow2; small pow3,5 and mix


bragadeesh:
Large1DThreshold cannot be same for single & double precision for performance reasons; 
The code that is in plan.cpp is the correct logic; maximum for single precision is 4096, and double precision is 2048; you need to rework other parts of code to fix bugs

tingxingdong:
The original logic fails in double

PR #75, opened 2017-11-07
remove static kernels of pow2

### LINKED ISSUE #67: Problems with double 1D complex interleaved for moderate to large sizes

tingxingdong:
resolves #67 

Summary of proposed changes:
-  remove expired pow2 kernels and their CPU functions. 
-  the large1D threadhold is stored in radix_table only instead of storing a variable per file.
-  double2 of pow2 split into multiple kernels starting at size 4096 (which is included). 
-  The cpu implementions of kernel_launch_single_large.cpp and kernel_launch_single.cpp are still sperate to speed up compilation. Too many kernels in one file prevent parallel compilation.


bragadeesh:
This PR still keeps the large 1d threshold as same between single and double; I was clear earlier that this is not what we want. Double max 1d must be half of single max 1d.
Can you rework the code?

tingxingdong:
You mean for pow2 only or for other pow3, 5?

bragadeesh:
What is Large1DThreshold for SP and DP in your code?
Its a common value, applicable to all radices. They don't vary across pow2 pow3 etc,

tingxingdong:
I only maintain one macro 4096.  The double one can be calculated accordingly. 
in plan.cpp 
 
size_t supported[] = {
                  4096, 4050, 4000, 3888, 3840, 3750, 3645, 3600, 3456, 3375, 3240, 3200, 3125, 3072, 3000, 2916, 2880, 2700, 2592, 2560, 2500, 2430, 2400, 2304, 2250, 2187, 2160,		                  4096, 4050, 4000, 3888, 3840, 3750, 3645, 3600, 3456, 3375, 3240, 3200, 3125, 3072, 3000, 2916, 2880, 2700, 2592, 2560, 2500, 2430, 2400, 2304, 2250, 2187, 2160,

If you want to drop pow3, 5 to 2K.  Then do you have to modify this line?

tingxingdong:
see line - 477 in plan.cpp.  if (length[0] > (Large1DThreshold * Large1DThreshold)) 

This is for non-pow2 case, Do you want to add an if-else branch to differentiate SP/DP or not?

PR #92, opened 2017-12-11
enable real FFT work 

tingxingdong:
Summary of proposed changes:
-  copy real buffer into complex buffer, a real FFT changes into a complex one 
-  add real FFT in tester
-  add descriptions in hipFFT API
-  pass on all single, pow2, 3, 5 and mix.


library/src/include/kernel_launch.h
```diff
@@ -19,8 +19,8 @@
 struct DeviceCallIn
 {
     TreeNode *node;
-    void *bufIn[2];
-    void *bufOut[2];
+    void *bufIn[2];//TODO, bufIn[1] is never used
+    void *bufOut[2];//TODO, bufOut[1] is never used
```
bragadeesh: actually when we support planar formats, the [1] indices would be used, just not now

library/src/powX.cpp
```diff
@@ -269,22 +308,23 @@ void TransformPowX(const ExecPlan &execPlan, void *in_buffer[], void *out_buffer
         hipMemcpy(dbg_out, data.bufOut[0], out_size_bytes, hipMemcpyDeviceToHost);
         printf("copied from device\n");
        
-        /*if(i == 0 || i == 2 || i == 4)
```
bragadeesh: why uncomment this code, this was just debug, we need to clean these up

PR #105, opened 2018-01-16
fill set_stream funciton 

tingxingdong:
-  add stream to every kernel launching
-  passed all the tests on rocm 1.7

### DIFF
--- a/library/src/include/transform.h
+++ b/library/src/include/transform.h
@@ -10,11 +10,11 @@
 
 struct rocfft_execution_info_t
 {
-	void *workBuffer;
-	size_t workBufferSize;
+    void *workBuffer;
+    size_t workBufferSize;
     hipStream_t rocfft_stream = 0;//by default it is stream 0
-	rocfft_execution_info_t() : workBuffer(nullptr), workBufferSize(0)
-	{}
+    rocfft_execution_info_t() : workBuffer(nullptr), workBufferSize(0)
+    {}
 };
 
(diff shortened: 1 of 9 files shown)


library/src/include/transform.h
```diff
@@ -6,13 +6,15 @@
...
 struct rocfft_execution_info_t
 {
-	void *workBuffer;
-	size_t workBufferSize;
-
-	rocfft_execution_info_t() : workBuffer(nullptr), workBufferSize(0)
-	{}
+    void *workBuffer;
+    size_t workBufferSize;
+    hipStream_t rocfft_stream = 0;//by default it is stream 0
```
bragadeesh: default must be initialized in the constructor initializer list

bragadeesh: hmm, i see c++ 11 allows this, never mind

tingxingdong:
I see the cross sign" The commit cannot be build". It is the jenkins machine down.

library/src/include/transform.h
tingxingdong: Yes, the rocBLAS also use this feature. 
https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/develop/library/src/include/handle.h#L31

bragadeesh:
@tingxingdong do you have some test code or example to show this works?

tingxingdong:
You mean,  set the stream? I think we can make one from the existing sample.

PR #123, opened 2018-04-20
fix building error on NV platform 

tingxingdong:
add a sample code to use multiple hip stream
-  
-  
-


saadrahim:
Can we close this request from April? I suggest reopening an issue if this is still a problem.

bragadeesh:
Merged the changes into cuda-build branch for now, can re-evaluate later

PR #127, opened 2018-06-19
Use hip device target when available

pfultz2:
This applies a hotfix, so rocFFT can build with hcc/hip from rocm 1.8.2.


bragadeesh:
thanks @pfultz2

PR #139, opened 2018-10-17
fix for case sensitive cmake to find HIP dependency

ntrost57:
resolves issues with case sensitive cmake to find HIP dependency (CentOS 7.5 w/ CMake 3.12)

### DIFF
--- a/library/src/CMakeLists.txt
+++ b/library/src/CMakeLists.txt
@@ -114,7 +114,7 @@ rocm_install_targets(
 rocm_export_targets(
   TARGETS roc::rocfft
   PREFIX rocfft
-  DEPENDS PACKAGE hip
+  DEPENDS PACKAGE HIP
   NAMESPACE roc::
  )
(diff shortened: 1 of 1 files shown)


library/src/CMakeLists.txt
```diff
@@ -114,7 +114,7 @@ rocm_install_targets(
 rocm_export_targets(
   TARGETS roc::rocfft
   PREFIX rocfft
-  DEPENDS PACKAGE hip
+  DEPENDS PACKAGE HIP
```
pfultz2: The package config file is name is [hip-config.cmake](https://github.com/ROCm-Developer-Tools/HIP/blob/master/CMakeLists.txt#L294) so it shouldn't be upper case as that would be the opposite of case sensitive.

ntrost57: Right, it should stay lower case. Thank you.

PR #141, opened 2018-11-03
hipFFT: R2C/C2R, streams, error code checking

### LINKED ISSUE #135: Status of hipFFT
What features of cuFFT are actually implemented?
It's hard to say looking at header and source files.

1. There is a lot of commented code and empty switch cases, for example: https://github.com/ROCmSoftwarePlatform/rocFFT/blob/develop/library/src/hipfft.cpp#L224

2. `hipfftExecC2C` calls the same code for different directions (unlike `hipfftExecZ2Z`):
https://github.com/ROCmSoftwarePlatform/rocFFT/blob/develop/library/src/hipfft.cpp#L858

3. Functions always return `HIPFFT_SUCCESS` and ignore result of `rocfft_*` functions.

4. Will hipFFT be a wrapper around cuFFT? Or it's supposed to be used with rocFFT only (on both ROCm and CUDA).

mkuron:
Resolves #135

Summary of proposed changes:
-  Add the missing cases in `hipfftMakePlan*` and `hipfftExec*` for R2C, C2R, D2Z, Z2D
-  Add the work buffer needed for this
-  Check rocFFT error codes and return hipFFT error codes
-  Implement `hipfftSetStream`

I have extensively tested the 3D case of R2C/C2R. Since the double-precision cases and the other dimensions have basically identical code, I expect them to work just as well.

### DIFF
--- a/library/src/hipfft.cpp
+++ b/library/src/hipfft.cpp
@@ -224,46 +224,139 @@ hipfftResult hipfftMakePlan1d(hipfftHandle plan,
     size_t lengths[1];
     lengths[0] = nx;
     size_t number_of_transforms = batch;
+    
+    size_t workBufferSize = 0;
 
     switch(type)
     {
         case HIPFFT_R2C:
-            return HIPFFT_NOT_IMPLEMENTED;
+            if(rocfft_plan_create_internal(plan->ip_forward,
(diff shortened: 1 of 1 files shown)


bragadeesh:
thanks @mkuron this is great

PR #142, opened 2018-11-14
make yum *install smoother

zaliu:
Summary of proposed changes:
-  add -y to yum localinstall to eliminate manual intervention
-  add --nogpgcheck to yum install to avoid failure when installing from unsigned repo


bragadeesh:
thanks @zaliu

PR #143, opened 2018-11-15
Refactoring hipfft

feizheng10:
Summary of proposed changes:
-  enabled versioning with git hash tag and added version string function
-  added err_check for each rocfft* and internal hipfft* calls in hipfft.cpp
-  combined duplicate code into one function hipfftMakePlan_internal() in hipfft.cpp
    (please double check the code at hipfft.cpp line:692-700 in pull request #141, is it specifically for 3d plan or works for all)
- changed a few return HIPFFT_SUCCESS to HIPFFT_NOT_IMPLEMENTED for non-implemented functions.
    (feel free to change them back...)


mkuron:
> please double check the code at hipfft.cpp line:692-700 in pull request #141, is it specifically for 3d plan or works for all

I assume you are referring to https://github.com/ROCmSoftwarePlatform/rocFFT/blob/b326977e3ed5f420f93927d3763003e0140e3341/library/src/hipfft.cpp#L692-L700

That is generic to all dimensions. I simply forgot to add it to the others... I think even `rocfft_plan_get_work_buffer_size` should be called for all four plans, not just for `op_forward`.

mkuron:
Why did this get closed? It fixes an actual bug by adding the work buffer to the non-3D transforms.

PR #145, opened 2018-11-23
dropping cmake3 in favor of cmake found on centos7

psteinb:
Summary of proposed changes:
-  dropping `cmake3` in favor of `cmake` because centos7 (7.4.1708 in my case) does not install `cmake3` from the repos

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -114,7 +114,7 @@ install_packages( )
 
   # dependencies needed for rocfft and clients to build
   local library_dependencies_ubuntu=( "make" "cmake-curses-gui" "hip_hcc" "pkg-config" )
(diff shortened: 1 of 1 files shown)


bragadeesh:
cmake3 is a requirement to build the library, are you not able to yum install cmake3?

psteinb:
Ah sorry my mistake, my centos7 install has EPEL and others enabled. 
This brought in a `cmake` binary which was 3.5 and hence the script 
didn't work.

psteinb:
you would wanna check the version of cmake and only take the one that yields cmake or higher. Quite some effort, when a more clean documentation of the cmake variables would do the trick.

PR #154, opened 2019-02-08
speed up build process

feizheng10:
Summary of proposed changes:
-  enable build switch -fno-gpu-rdc for rocFFT device lib
-  split small kernels into groups


feizheng10:
In the original code, 
  function_pool.cpp 
  kernel_launch_single.cpp 
  kernel_launch_double.cpp 
  kernel_launch_single_large.cpp 
  kernel_launch_double_large.cpp
are the entries for the relevant *.h. All the *.cpp are checked in, and all the *.h are generated in build directory with generated kernels.
With build profiling, we found kernel_launch_single.cpp and kernel_launch_double.cpp are the compiling bottleneck because all small kernels go into single cpp file.
Now, we could make small kernels into groups, and each group goes into 1 cpp/h file pair. And then we could take advantage of parallel compiling more. The total small kernel group number can be configured in library/src/device/generator/generated-kernels.cmake.
As well, configuring in generated-kernels.cmake, we could generate the above *.cpp in build directory instead of checking in the repo. The generated files are still the output of the custom build command in /library/src/device/CMakeLists.txt. Finally, the generated files and explicit specified source files are the source of fft device library.

PR #171, opened 2019-05-03
Enable --prefix option for install.sh

malcolmroberts:
Only elevate permission is needed in install.sh
Indent to 4 spaces and wrap at 100 lines in install.sh

Summary of proposed changes:
-  enable --prefix option for install.sh
-  only elevate make install if required
-  indent to 4 spaces, wrap at 100 columns

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -8,159 +8,189 @@
 
 function display_help()
 {
-  echo "rocfft build & installation helper script"
-  echo "./install [-h|--help] "
-  echo "    [-h|--help] prints this help message"
-#  echo "    [--prefix] Specify an alternate CMAKE_INSTALL_PREFIX for cmake"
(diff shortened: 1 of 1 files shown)


yxsamliu:
This change seems to cause install.sh -icd not working for us.
If I execute `install.sh -icd`, it will output the following then exit
`./install.sh -icd
Creating project build directory in: ./build
make cmake-curses-gui hip_hcc pkg-config
libfftw3-dev libboost-program-options-dev
`

malcolmroberts:
Looks like I left in some debugging code accidentally; sorry about that.

I have a fix in https://github.com/malcolmroberts/rocFFT/tree/install_prefix (0931b6f) and we'll try to get that in develolp quickly.

PR #173, opened 2019-05-07
Packaging clients

feizheng10:
Summary of proposed changes:
This is a simple solution to package clients with dpkg. And also it is portable for other libs.


malcolmroberts:
The debug build fails with the PrintNode issue; do you need to rebase on develop?

feizheng10:
good call

malcolmroberts:
Looks good.

PR #175, opened 2019-05-08
removed boost dependency of hipfft

feizheng10:
removed boost dependency of hipfft


feizheng10:
The CI failure came from clang-formatter

PR #178, opened 2019-05-09
Remove accidentally added debugging code in install.sh

malcolmroberts:
Summary of proposed changes:
-  Remove accidentally added debugging code in install.sh which broke ./install.sh -icd


feizheng10:
The CI failure came from one uncompleted configured  CI node. All other tests on other nodes are fine.

PR #183, opened 2019-05-16
Clean up 1D plan generation

malcolmroberts:
Summary of proposed changes:
-  Break up the 1D plan generation into separate functions.

### DIFF
--- a/library/src/include/tree_node.h
+++ b/library/src/include/tree_node.h
@@ -8,6 +8,7 @@
 
 #include <cstring>
 #include <iostream>
+#include <map>
 #include <vector>
 
 #include "kargs.h"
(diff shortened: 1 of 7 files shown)


library/src/include/tree_node.h
```diff
@@ -92,8 +93,32 @@ class TreeNode
...
+        Pow2Lengths1Double.insert(std::make_pair(16384, 64));
+        Pow2Lengths1Double.insert(std::make_pair(32768, 128));
+        Pow2Lengths1Double.insert(std::make_pair(65536, 64));
+        Pow2Lengths1Double.insert(std::make_pair(131072, 64));
     }
 
+    // Maps from length[0] to divLength1 for 1D transforms in
+    // single and double precision for power-of-two transfor sizes
+    // using blocks.
+    std::map<size_t, size_t> Pow2Lengths1Single;
```
feizheng10: Good for now.  I still think subclass for each instance of float and double would be better.

PR #190, opened 2019-06-03
Multigpu

akilaMD:
- Ignore for now. This is just here to test on multi-gpu node


holyprince:
Sorry to interrupt you, I looked the commit and want to ask whether the FFT library support multiGPU ?
 I have large data to do FFT on GPU and want to use multi GPU.

PR #204, opened 2019-06-20
Implement efficient real/complex 1D even-length transforms

malcolmroberts:
Summary of proposed changes:
-  Implement time and memory efficient real/complex 1D transforms for even problem sizes.
-  Includes some general code cleaning.


clients/tests/hipfft_test.cpp
```diff
@@ -89,7 +103,14 @@ TEST(hipfftTest, CheckBufferSizeD2Z)
...
 
     EXPECT_TRUE(hipfftMakePlan1d(plan, n, HIPFFT_D2Z, batch, &workSize) == HIPFFT_SUCCESS);
-    EXPECT_TRUE(workSize == 2 * n * batch * sizeof(double));
+    if(n % 2 == 0)
+    {
+        EXPECT_TRUE(workSize == 0);
+    }
+    else
+    {
+        EXPECT_TRUE(workSize == 2 * n * sizeof(float));
```
feizheng10: double?

clients/tests/hipfft_test.cpp
```diff
@@ -103,7 +124,14 @@ TEST(hipfftTest, CheckBufferSizeZ2D)
...
 
     EXPECT_TRUE(hipfftMakePlan1d(plan, n, HIPFFT_Z2D, batch, &workSize) == HIPFFT_SUCCESS);
-    EXPECT_TRUE(workSize == 2 * n * batch * sizeof(double));
+    if(n % 2 == 0)
+    {
+        EXPECT_TRUE(workSize == 0);
+    }
+    else
+    {
+        EXPECT_TRUE(workSize == 2 * n * sizeof(float));
```
feizheng10: double?

clients/tests/hipfft_test.cpp
malcolmroberts: Thanks; fixed in 0cfc23b.

clients/tests/hipfft_test.cpp
malcolmroberts: Thanks; fixed in 0cfc23b.

library/src/plan.cpp
```diff
@@ -192,6 +265,12 @@ rocfft_status rocfft_plan_create_internal(rocfft_plan                   plan,
             {
                 if(description->inArrayType != description->outArrayType)
                     return rocfft_status_invalid_array_type;
+                if(description->inStrides != description->outStrides)
```
feizheng10: double check this, I thought we removed it.

malcolmroberts: The type has been changed from size_t* to std::array<size_t, 3>, so the check is on the data, not the address.  However, this isn't the correct place to check this, so we're probably just looking at the default values, so I removed it in d91b984.

library/src/plan.cpp
```diff
@@ -1744,99 +1944,119 @@ void TreeNode::TraverseTreeAssignBuffersLogicA(OperatingBuffer& flipIn,
...
+                if(parent->iOffset != 0)
+                    iOffset = parent->iOffset;
+                break;
+            default:
+                inArrayType = rocfft_array_type_complex_interleaved;
+            }
         }
 
-        switch(obOut)
+        if(outArrayType == rocfft_array_type_unset)
```
feizheng10: we might remove redundant code if the logic for in/out are the same later.

akilaMD:
So this is still failing. I say we merge it in for now and deal with the librocfft linking issues ASAP.

library/include/rocfft.h
```diff
@@ -182,10 +183,10 @@ ROCFFT_EXPORT rocfft_status rocfft_plan_create(rocfft_plan*                  pla
...
  * rocfft_execution_info_create
  *  */
-ROCFFT_EXPORT rocfft_status rocfft_execute(const rocfft_plan     plan,
-                                           void*                 in_buffer[],
-                                           void*                 out_buffer[],
-                                           rocfft_execution_info info);
+ROCFFT_EXPORT rocfft_status rocfft_execute(const rocfft_plan           plan,
+                                           void*                       in_buffer[],
+                                           void*                       out_buffer[],
+                                           const rocfft_execution_info info);
```
bragadeesh: info cannot be const, I think it was intended to act as both an input & output parameter, the library may be returning information back to user

library/include/rocfft.h
```diff
@@ -210,54 +211,48 @@ ROCFFT_EXPORT rocfft_status rocfft_plan_description_set_scale_float( rocfft_plan
 ROCFFT_EXPORT rocfft_status rocfft_plan_description_set_scale_double( rocfft_plan_description description, double scale );
```
bragadeesh: change the scale to const here as well to be consistent?

library/include/rocfft.h
malcolmroberts: Sounds good; done.

library/include/rocfft.h
malcolmroberts: Sounds good; done.

PR #206, opened 2019-06-20
[rocfft] Include <iostream> as `std::cerr` is referenced.

mhbliao:
It is required to fix TC build as reported in http://ontrack-internal.amd.com/browse/SWDEV-193409

### DIFF
--- a/library/src/auxiliary.cpp
+++ b/library/src/auxiliary.cpp
@@ -27,6 +27,7 @@
 #include "logging.h"
 #include "rocfft.h"
 #include "rocfft_hip.h"
+#include <iostream>
 
 /*******************************************************************************
  * Static handle data


mhbliao:
> This breaks the Jenkins build, which currently uses rocm's clang-format, and we can't move forward with this PR until this is resolved.
> 
> Please use clang-format on the modified file, which will move the iostream include the end of that particular block of #includes (there is alphabetical ordering going on).

done in new change

mhbliao:
BTW, for unknown reason, that git hook is not triggered in my local environment, sound weird.

malcolmroberts:
Applying clang-format isn't a git hook (though that might be a good idea), but it's in the Jenkins file.  There's currently the issue that the clang-format in rocm2.4 gives different results than that in rocm2.5, but once this is stable we might make this a git hook.

PR #207, opened 2019-06-21
Added hip-clang support for rocfft

amdkila:
- Given that we're switching to the hip-clang compiler very soon, we need to add a 3rd CI job that tests the hip-clang compiler
- Passed CI


amdkila:
After 5 weeks this is finally passing CI! Let's merge this

PR #212, opened 2019-07-04
Remove FFTW dependency for hipfft examples and simplify examples.

malcolmroberts:
Summary of proposed changes:
-  Removes FFTW dependency for hipfft examples
-  Simplify examples so that they're more pedagogical.

### DIFF
--- a/clients/samples/hipfft/hipfft_planmany_2d_r2c.cpp
+++ b/clients/samples/hipfft/hipfft_planmany_2d_r2c.cpp
@@ -6,6 +6,8 @@
 
 int main()
 {
+    std::cout << "2D double-precision real-to-complex transform using advanced interface\n";
+    
     int rank    = 2;
     int n[2]    = {4, 5};
     int howmany = 3; // batch size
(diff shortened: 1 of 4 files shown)


clients/samples/hipfft/hipfft_planmany_2d_r2c.cpp
```diff
@@ -1,103 +1,67 @@
...
+#include <vector>
 
 int main()
 {
-    ///------------------------------------------------------------------------
-    /// fftwf_plan_many_dft_r2c vs hipfftPlanMany
-    /// 2D R2C forward in-place
+    int rank    = 2;
+    int n[2]    = {4, 5};
+    int howmany = 3; // batch size
```
feizheng10: It is better to leave the comment "forward in-place"

malcolmroberts: Commit 6e28ab0 adds console output which also serves as a comment (also for z2z case).

PR #213, opened 2019-07-04
improve perf test

feizheng10:
* added full test suite in perf test
* saved test config info to file


malcolmroberts:
When one runs ./rocfft_perf_test_suite.py without arguments one gets a python error; this used to return

Error! Can not find ref file ./short_f0_ref.csv

PR #217, opened 2019-07-11
Add more hipfft examples

malcolmroberts:
Summary of proposed changes:
-  Add hipfft examples for 1D, 2D, and 3D, complex-complex and real-complex transforms using the simple interface.
- Use std::complex in rocfft example (allowing for simpler output).


clients/samples/hipfft/hipfft_1d_d2z.cpp
```diff
@@ -0,0 +1,109 @@
...
+* AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+* LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+* OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+* THE SOFTWARE.
+*******************************************************************************/
+
+#include <complex>
+#include <iostream>
+
+#include <hip/hip_runtime.h>
```
feizheng10: change all the hip_runtime.h to hip_runtime_api.h for hcc future release

clients/samples/hipfft/hipfft_1d_d2z.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include <hipfft.h>
+
+// Kernel for initializing the real-valued input data on the GPU.
+__global__ void initdata(double* x, const int Nx)
```
feizheng10: Not sure we should put all these functions in 1 common file

clients/samples/hipfft/hipfft_planmany_2d_r2c.cpp
```diff
@@ -6,7 +6,7 @@
 
 int main()
 {
-    std::cout << "2D double-precision real-to-complex transform using advanced interface\n";
+    std::cout << "hipfft 2D double-precision real-to-complex transform using advanced interface\n";
```
feizheng10: single-precision

clients/samples/hipfft/hipfft_1d_d2z.cpp
malcolmroberts: They're different for the different cases, and I find it more pedagogical to have them in the same source tree.

clients/samples/hipfft/hipfft_planmany_2d_r2c.cpp
malcolmroberts: Done; thanks.

clients/samples/hipfft/hipfft_1d_d2z.cpp
malcolmroberts: If I do this, then compilation fails with hipcc; let's discuss.

malcolmroberts: Seems to work if one doesn't actually have kernel code in the compilation unit.

PR #241, opened 2019-08-23
Fix index 0 calculation c2r prekernel to deal with non-zero imaginary part

malcolmroberts:
The real-to-complex pre-processing kernel incorrectly assumed that the first and last index had
zero imaginary part, which may not be true when used to pre-process or 2D and 3D transforms.

Summary of proposed changes:
-  Account for imaginary part for index-0 computations in real_1d_pre_post_process_kernel


malcolmroberts:
Closed in favour of 248.

PR #243, opened 2019-08-26
Making rocFFT SLES ready

adilad:
Resolves the issue of rocFFT not building on SLES

Summary of proposed changes:
-  This makes sure that rocFFT can be built on SLES systems
-  
-


amdkila:
See PR #245

PR #245, opened 2019-08-28
Adding sles to CI and added some missing packages

amdkila:
- Need to merge in rocJenkins PR first before testing this in CI
- PR #243 contains some incorrect packages in the install script. They're fixed here

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -35,7 +35,7 @@ rocFFTCI:
     rocfft.paths.build_command = './install.sh -c'
 
     // Define test architectures, optional rocm version argument is available
-    def nodes = new dockerNodes(['gfx900 && ubuntu', 'gfx906 && centos', 'gfx906 && sles', 'gfx900 && centos7 && hip-clang','gfx906 && ubuntu && hip-clang'], rocfft)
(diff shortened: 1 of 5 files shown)


saadrahim:
> 
> 
> Looks good; a couple of comments need to be updated.
> 
> In the future we might consider just providing cmake and removing install.sh.

CI is heavily dependent on install.sh. Can we just move it to .jenkins folder?

Jenkinsfile
```diff
@@ -71,24 +71,11 @@ rocFFTCI:
...
-        
-        if(platform.jenkinsLabel.contains('centos'))
-        {
-            command = """#!/usr/bin/env bash
+        def command = """#!/usr/bin/env bash
                     set -x
                     cd ${project.paths.project_build_prefix}/build/release/clients/staging
-                    LD_LIBRARY_PATH=/opt/rocm/hcc/lib GTEST_LISTENER=NO_PASS_LINE_IN_LOG sudo ./rocfft-test --gtest_output=xml --gtest_color=yes
+                    LD_LIBRARY_PATH=/opt/rocm/lib GTEST_LISTENER=NO_PASS_LINE_IN_LOG sudo ./rocfft-test --gtest_output=xml --gtest_color=yes
                 """
```
saadrahim: Why is the Centos fix here if we are working on SLES?

PR #248, opened 2019-09-03
Efficient 2D and 3D complex-to-real transforms for even-length contiguous dimension

malcolmroberts:
[2dc2reven_mi60.pdf](https://github.com/ROCmSoftwarePlatform/rocFFT/files/3570725/2dc2reven_mi60.pdf)
[2dc2reven_vega20.pdf](https://github.com/ROCmSoftwarePlatform/rocFFT/files/3570727/2dc2reven_vega20.pdf)
[2dc2reven_vega10.pdf](https://github.com/ROCmSoftwarePlatform/rocFFT/files/3570729/2dc2reven_vega10.pdf)

Summary of proposed changes:
-  Implement efficient 2D and 3D complex-to-real transforms when fastest dimension is even.
-  Disable a 2D test case where the (unscaled) error is slightly too high with the new code.

Benchmark figures comparing with develop (92c4537652cc90d68d96af60962bfb1c3c870811) are attached.


clients/tests/accuracy_test_2D.cpp
```diff
@@ -58,12 +58,15 @@ class accuracy_test_complex_2D_pow2_double : public ::testing::Test
     virtual void TearDown() {}
 };
 // 65536=pow(2,16)                                 //8388608 = pow(2,23)
-#define POW2_RANGE                                        \
-    {2, 4}, {8, 16}, {32, 128}, {256, 512}, {1024, 2048}, \
-    {                                                     \
-        4096, 8192                                        \
-    } /* malloc fail on 4GB Fiji Nano on the following size \                    \
-, {16384, 32768}, {65536, 131072}, {262144, 524288} */
+#define POW2_RANGE                          \
```
feizheng10: POW2_RANGE seems to be shared for regular C2C cases also. The better way could be: define a sub-range just for R2C cases.

malcolmroberts: Sounds reasonable; done.

PR #269, opened 2019-11-14
rocFFT - Changes to support multiple ROCM installation

pruthvistony:
- Hard coded reference to /opt/rocm is parameterized
  to CMAKE_INSTALL_PREFIX
- Package is generated to install into ROCM_INSTALL_PATH if
  defined in the env otherwise default into /opt/rocm
- RUNPATH is set in the lib only if ROCM_RPATH env is defined
  with the rpath

Signed-off-by: Pruthvi Madugundu <mpruthvi@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,13 +93,13 @@ elseif( CMAKE_CXX_COMPILER MATCHES ".*/hipcc$" )
   set( CMAKE_C_COMPILE_OPTIONS_VISIBILITY_INLINES_HIDDEN "-Xcompiler ${CMAKE_C_COMPILE_OPTIONS_VISIBILITY_INLINES_HIDDEN}" )
   set( CMAKE_CXX_COMPILE_OPTIONS_VISIBILITY_INLINES_HIDDEN "-Xcompiler ${CMAKE_CXX_COMPILE_OPTIONS_VISIBILITY_INLINES_HIDDEN}" )
 elseif( CMAKE_CXX_COMPILER MATCHES ".*/hcc$" )
-  message( STATUS "HCC compiler set; ROCm backend selected [ CXX=/opt/rocm/bin/hcc cmake ... ]" )
+  message( STATUS "HCC compiler set; ROCm backend selected [ CXX=/opt/rocm-<ver>/bin/hcc cmake ... ]" )
 endif( )
 
 # This finds the rocm-cmake project, and installs it if not found
 # rocm-cmake contains common cmake code for rocm projects to help setup and install
 set( PROJECT_EXTERN_DIR ${CMAKE_CURRENT_BINARY_DIR}/extern )
-find_package( ROCM CONFIG QUIET PATHS /opt/rocm )
+find_package( ROCM CONFIG QUIET PATHS ${CMAKE_INSTALL_PREFIX} )
 if( NOT ROCM_FOUND )
   set( rocm_cmake_tag "master" CACHE STRING "rocm-cmake tag to download" )
   file( DOWNLOAD https://github.com/RadeonOpenCompute/rocm-cmake/archive/${rocm_cmake_tag}.zip
@@ -126,7 +126,7 @@ rocm_setup_version( VERSION ${VERSION_STRING} )
 list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake )
 
 # NOTE:  workaround until hcc & hip cmake modules fixes symlink logic in their config files; remove when fixed
-list( APPEND CMAKE_PREFIX_PATH /opt/rocm/hcc /opt/rocm/hip )
+list( APPEND CMAKE_PREFIX_PATH ${CMAKE_INSTALL_PREFIX}/hcc ${CMAKE_INSTALL_PREFIX}/hip )
 
 # Enable verbose output
 option( BUILD_VERBOSE "Output additional build information" OFF )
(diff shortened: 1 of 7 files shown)


install.sh
```diff
@@ -269,6 +269,10 @@ build_clients=false
 build_cuda=false
 build_release=true
 
+if ! [ -z ${ROCM_INSTALL_PATH+x} ]; then
+    install_prefix=${ROCM_INSTALL_PATH}
+fi
```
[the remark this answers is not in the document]
malcolmroberts: I'm more in favour of deprecating install.sh and using cmake.  This will require some change in CI, however.

cmake/package-functions.cmake
```diff
@@ -11,7 +11,7 @@ file( WRITE ${scripts_write_dir}/postinst
 set -e
 
 do_ldconfig() {
-  echo ${CPACK_PACKAGING_INSTALL_PREFIX}/${LIB_INSTALL_DIR} > ${ld_conf_file} && ldconfig
+  echo /opt/rocm/${LIB_INSTALL_DIR} > ${ld_conf_file} && ldconfig
```
feizheng10: Could you explain the change a little bit? Why it is going to /opt/rocm/?

CMakeLists.txt
```diff
@@ -93,13 +93,13 @@ elseif( CMAKE_CXX_COMPILER MATCHES ".*/hipcc$" )
   set( CMAKE_C_COMPILE_OPTIONS_VISIBILITY_INLINES_HIDDEN "-Xcompiler ${CMAKE_C_COMPILE_OPTIONS_VISIBILITY_INLINES_HIDDEN}" )
   set( CMAKE_CXX_COMPILE_OPTIONS_VISIBILITY_INLINES_HIDDEN "-Xcompiler ${CMAKE_CXX_COMPILE_OPTIONS_VISIBILITY_INLINES_HIDDEN}" )
 elseif( CMAKE_CXX_COMPILER MATCHES ".*/hcc$" )
-  message( STATUS "HCC compiler set; ROCm backend selected [ CXX=/opt/rocm/bin/hcc cmake ... ]" )
+  message( STATUS "HCC compiler set; ROCm backend selected [ CXX=/opt/rocm-<ver>/bin/hcc cmake ... ]" )
```
saadrahim: This line looks broken, how will it work if the user wants to use /opt/rocm/ only?

CMakeLists.txt
```diff
@@ -93,13 +93,13 @@ elseif( CMAKE_CXX_COMPILER MATCHES ".*/hipcc$" )
...
 endif( )
 
 # This finds the rocm-cmake project, and installs it if not found
 # rocm-cmake contains common cmake code for rocm projects to help setup and install
 set( PROJECT_EXTERN_DIR ${CMAKE_CURRENT_BINARY_DIR}/extern )
-find_package( ROCM CONFIG QUIET PATHS /opt/rocm )
+find_package( ROCM CONFIG QUIET PATHS ${CMAKE_INSTALL_PREFIX} )
```
saadrahim: CMAKE_INSTALL_PREFIX is where the library should be installed. It is not the root directory where the library should search for dependencies.

install.sh
```diff
@@ -368,6 +372,9 @@ fi
 # We append customary rocm path; if user provides custom rocm path in ${path}, our
 # hard-coded path has lesser priority
 export PATH=${PATH}:/opt/rocm/bin
+if ! [ -z ${ROCM_PATH+x} ]; then
+    export PATH=${PATH}:${ROCM_PATH}/bin
```
saadrahim: This is appending the ROCM_PATH to the end, it is useless as /opt/rocm/ is supposed to be available as a software link and always exists.

malcolmroberts:
CMAKE_INSTALL_PREFIX is the install path; this shouldn't be used to specify where the include and linked libraries are stored.  I've been using this to set where to install for months; users will expect this to work as per the cmake documentation as well.

malcolmroberts:
Once these issues are resolved we can re-open the PR.

install.sh
pruthvistony: The symlink is needed only when user is using frameworks like pytorch/Caffe2 or any benchmark framework. It helps to keep an uniform interface when single/multiple ROCm versions are installed on a machine. 
In the above scenario while compiling the rocFFT it is required to point to the exact path of ROCm so that we pick the correct version of compiler/hcc binary.

CMakeLists.txt
pruthvistony: As I updated in the other comment, the symlink is needed only when user is using frameworks like pytorch/Caffe2 or any benchmark framework. It helps to keep an uniform interface when single/multiple ROCm versions are installed on a machine.
In the above scenario while compiling the rocFFT it is required to point to the exact path of ROCm so that we pick the correct version of compiler/hcc binary.

cmake/package-functions.cmake
pruthvistony: This change is not required. Since itseems the function write_rocm_package_script_files() is not used/called. However the corresponding change is in library/CMakelists.txt - ROCFFT_CONFIG_DIR is passed to rocm_package_create(). So reverting this change.

install.sh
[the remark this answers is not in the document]
pruthvistony: install_prefix is set to CMAKE_INSTALL_PREFIX, it is set a default value of rocfft-install, which I believe is set from the CI build. So want to overwrite it only if ROCM_PATH is set.

install.sh
pruthvistony: As already mentioned, symlink doesn't always exist.

CMakeLists.txt
pruthvistony: Updated all the changes to remove the CMAKE_INSTALL_PREFIX and raised a new PR
https://github.com/ROCmSoftwarePlatform/rocFFT/pull/275
Please review

pruthvistony:
Updated all the changes to remove the CMAKE_INSTALL_PREFIX and raised a new PR
https://github.com/ROCmSoftwarePlatform/rocFFT/pull/275
Please review

PR #275, opened 2019-12-03
rocFFT Changes to support multiple ROCM installation

pruthvistony:
- New mode of building is added "-r,--relocatable" which is used
  for ROCm stack installed in /opt/rocm-ver.
- Below CMAKE parameters are set/overwritten in above mode
  CMAKE_INSTALL_PREFIX
  CPACK_PACKAGING_INSTALL_PREFIX
  CMAKE_PREFIX_PATH
  CMAKE_SHARED_LINKER_FLAGS
  ROCM_DISABLE_LDCONFIG

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,6 +96,9 @@ elseif( CMAKE_CXX_COMPILER MATCHES ".*/hcc$" )
   message( STATUS "HCC compiler set; ROCm backend selected [ CXX=/opt/rocm/bin/hcc cmake ... ]" )
 endif( )
 
+# NOTE:  workaround until hcc & hip cmake modules fixes symlink logic in their config files; remove when fixed
+list( APPEND CMAKE_PREFIX_PATH /opt/rocm /opt/rocm/hcc /opt/rocm/hip )
+
 # This finds the rocm-cmake project, and installs it if not found
 # rocm-cmake contains common cmake code for rocm projects to help setup and install
 set( PROJECT_EXTERN_DIR ${CMAKE_CURRENT_BINARY_DIR}/extern )
@@ -118,15 +121,12 @@ include( ROCMPackageConfigHelpers )
 include( ROCMInstallSymlinks )
 
 # Using standardized versioning from rocm-cmake
-set ( VERSION_STRING "0.9.7" )
+set ( VERSION_STRING "0.9.8" )
 rocm_setup_version( VERSION ${VERSION_STRING} )
 
 # Append our library helper cmake path and the cmake path for hip (for convenience)
 # Users may override HIP path by specifying their own in CMAKE_MODULE_PATH
 list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake )
-
-# NOTE:  workaround until hcc & hip cmake modules fixes symlink logic in their config files; remove when fixed
-list( APPEND CMAKE_PREFIX_PATH /opt/rocm/hcc /opt/rocm/hip )
 
 # Enable verbose output
 option( BUILD_VERBOSE "Output additional build information" OFF )
--- a/install.sh
+++ b/install.sh
@@ -36,6 +36,7 @@ function display_help()
     echo "    [-d|--dependencies] install build dependencies"
     echo "    [-c|--clients] build library clients too (combines with -i & -d)"
     echo "    [-g|--debug] -DCMAKE_BUILD_TYPE=Debug (default is =Release)"
+    echo "    [-r]--relocatable] create a package to support relocatable ROCm"
     echo "    [--cuda] build library for cuda backend"
     echo "    [--hip-clang] build library for amdgpu backend using hip-clang"
 }
@@ -268,6 +269,7 @@ install_prefix=rocfft-install
 build_clients=false
 build_cuda=false
 build_release=true
+build_relocatable=false
 
 # #################################################
 # Parameter parsing
@@ -276,7 +278,7 @@ build_release=true
 # check if we have a modern version of getopt that can handle whitespace and long parameters
 getopt -T
 if [[ $? -eq 4 ]]; then
-    GETOPT_PARSE=$(getopt --name "${0}" -o 'hidcg' --long 'help,install,clients,dependencies,debug,cuda,hip-clang,prefix:' --options hicgd -- "$@")
+    GETOPT_PARSE=$(getopt --name "${0}" -o 'hidcgr' --long 'help,install,clients,dependencies,debug,cuda,hip-clang,prefix:,relocatable' --options hicgdr -- "$@")
 else
     echo "Need a new version of getopt"
     exit 1
@@ -303,6 +305,9 @@ while true; do
             shift ;;
         -c|--clients)
             build_clients=true
+            shift ;;
+        -r|--relocatable)
+            build_relocatable=true
             shift ;;
         -g|--debug)
             build_release=false
@@ -337,6 +342,18 @@ else
     rm -rf ${build_dir}/debug
 fi
 
+if [[ "${build_relocatable}" == true ]]; then
+    rocm_path=/opt/rocm
+    if ! [ -z ${ROCM_PATH+x} ]; then
+        rocm_path=${ROCM_PATH}
+    fi
+
+    rocm_rpath=" -Wl,--enable-new-dtags -Wl,--rpath,/opt/rocm/lib:/opt/rocm/lib64"
+    if ! [ -z ${ROCM_RPATH+x} ]; then
+        rocm_rpath=" -Wl,--enable-new-dtags -Wl,--rpath,${ROCM_RPATH}"
+    fi
+fi
+
 # Default cmake executable is called cmake
 cmake_executable=cmake
 
@@ -367,7 +384,11 @@ fi
 
 # We append customary rocm path; if user provides custom rocm path in ${path}, our
 # hard-coded path has lesser priority
-export PATH=${PATH}:/opt/rocm/bin
+if [[ "${build_relocatable}" == true ]]; then
+    export PATH=${rocm_path}/bin:${PATH}
+else
+    export PATH=${PATH}:/opt/rocm/bin
+fi
 
 pushd .
 # #################################################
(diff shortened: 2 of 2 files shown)


malcolmroberts:
I had a look at the code, and I managed to create a branch with no reference to /opt/rocm in the cmake files.  This is available here: https://github.com/malcolmroberts/rocFFT/tree/no_opt_rocm

I was able to compile and install with environment variables telling cmake where to look for hcc, etc.  I didn't use the install script; one can simply use cmake, via 

```
cmake -DCMAKE_CXX_COMPILER=hcc -DBUILD_CLIENTS_TESTS=on -DCMAKE_INSTALL_PREFIX=${HOME}/rocfft ..
```

which sets the compiler, sets the install directory to ~/rocfft, and turns on building the tests.  This seems like a solution which doesn't require significant changes.

I'd have to look through my .bashrc to determine which variables are necessary; I suspect that it's just

```
export HCC_ROOT=/opt/rocm/hcc
if [ -v HCC_ROOT ]; then
    export PATH=${HCC_ROOT}/bin:${PATH}
    export HCC_INCLUDE_PATH=${HCC_ROOT}/include
    export CPLUS_INCLUDE_PATH=${HCC_INCLUDE_PATH}:${CPLUS_INCLUDE_PATH}
    export CMAKE_PREFIX_PATH=${HCC_HOME}/lib/cmake/hcc:${CMAKE_PREFIX_PATH}
fi
```

Since requires very minimal changes to cmake, and doesn't add extra variables, it seems preferable.  Would this be an acceptable solution for 10.1?

CMakeLists.txt
```diff
@@ -93,13 +93,13 @@ elseif( CMAKE_CXX_COMPILER MATCHES ".*/hipcc$" )
   set( CMAKE_C_COMPILE_OPTIONS_VISIBILITY_INLINES_HIDDEN "-Xcompiler ${CMAKE_C_COMPILE_OPTIONS_VISIBILITY_INLINES_HIDDEN}" )
   set( CMAKE_CXX_COMPILE_OPTIONS_VISIBILITY_INLINES_HIDDEN "-Xcompiler ${CMAKE_CXX_COMPILE_OPTIONS_VISIBILITY_INLINES_HIDDEN}" )
 elseif( CMAKE_CXX_COMPILER MATCHES ".*/hcc$" )
-  message( STATUS "HCC compiler set; ROCm backend selected [ CXX=/opt/rocm/bin/hcc cmake ... ]" )
+  message( STATUS "HCC compiler set; ROCm backend selected [ CXX=/opt/rocm-<ver>/bin/hcc cmake ... ]" )
```
feizheng10: This is still hard-coded.

CMakeLists.txt
```diff
@@ -93,13 +93,13 @@ elseif( CMAKE_CXX_COMPILER MATCHES ".*/hipcc$" )
...
 endif( )
 
 # This finds the rocm-cmake project, and installs it if not found
 # rocm-cmake contains common cmake code for rocm projects to help setup and install
 set( PROJECT_EXTERN_DIR ${CMAKE_CURRENT_BINARY_DIR}/extern )
-find_package( ROCM CONFIG QUIET PATHS /opt/rocm )
+find_package( ROCM CONFIG QUIET PATHS ${ROCM_PATH} /opt/rocm )
```
feizheng10: The clear and better way to use ${ROCM_PATH} is:
check ROCM_PATH has been set by user or not, if not set it to /opt/rocm only once in the root CMakelist.
As well, message the information of the ROCM_PATH set by user or not and the exact path.
And remove all the below "/opt/rocm" from any search path.

malcolmroberts: Moreover, ROCM_PATH is unnecessary; we can just use CMAKE_PREFIX_PATH.

library/src/CMakeLists.txt
```diff
@@ -74,6 +74,12 @@ prepend_path( ".." rocfft_headers_public relative_rocfft_headers_public )
 
 add_library( rocfft ${rocfft_source} ${relative_rocfft_headers_public} )
 add_library( roc::rocfft ALIAS rocfft )
+
+# RUNPATH is set only when ROCM_RPATH is defined in the ENV
+if( DEFINED ENV{ROCM_RPATH} )
+  set ( CMAKE_SHARED_LINKER_FLAGS " -Wl,--enable-new-dtags -Wl,--rpath,$ENV{ROCM_RPATH}" )
```
feizheng10: If this works, I believe we need this for rocfft-device build as well.

pruthvistony:
> 
> 
> I had a look at the code, and I managed to create a branch with no reference to /opt/rocm in the cmake files. This is available here: https://github.com/malcolmroberts/rocFFT/tree/no_opt_rocm
> 
> I was able to compile and install with environment variables telling cmake where to look for hcc, etc. I didn't use the install script; one can simply use cmake, via
> 
> ```
> cmake -DCMAKE_CXX_COMPILER=hcc -DBUILD_CLIENTS_TESTS=on -DCMAKE_INSTALL_PREFIX=${HOME}/rocfft ..
> ```
> 
> which sets the compiler, sets the install directory to ~/rocfft, and turns on building the tests. This seems like a solution which doesn't require significant changes.
> 
> I'd have to look through my .bashrc to determine which variables are necessary; I suspect that it's just
> 
> ```
> export HCC_ROOT=/opt/rocm/hcc
> if [ -v HCC_ROOT ]; then
>     export PATH=${HCC_ROOT}/bin:${PATH}
>     export HCC_INCLUDE_PATH=${HCC_ROOT}/include
>     export CPLUS_INCLUDE_PATH=${HCC_INCLUDE_PATH}:${CPLUS_INCLUDE_PATH}
>     export CMAKE_PREFIX_PATH=${HCC_HOME}/lib/cmake/hcc:${CMAKE_PREFIX_PATH}
> fi
> ```
> 
> Since requires very minimal changes to cmake, and doesn't add extra variables, it seems preferable. Would this be an acceptable solution for 10.1?

Hi Malcolm Roberts,

I agree this implementation also works. Adding of ROCM_PATH is done so as to provide a uniform way of referring to all components of ROCm stack at any custom path. In the case of rocFFT only hcc is required from ROCm stack, but while building other components, more dependencies are needed to be fulfilled. 
I believe the changes look more due to more files involved, I can remove few other by removing the changes in comments.

Regarding the changes for setting a hard coded path for ld.conf is not yet finalized. I will get back on this part as early as possible.

library/src/CMakeLists.txt
pruthvistony: This works, we have made a complete ROCm 2.10.1 build based on these changes. QA tested all the cases and approved the build also.

CMakeLists.txt
pruthvistony: @feizheng10 ,
Default value of ${ROCM_PATH} is not set as it is not required, if not defined/set the search path will include only /opt/rocm. 

@malcolmroberts,
Reason for not adding into CMAKE_PREFIX_PATH was to give highest priority to the user provided ROCm path as part of the find_package parameter, so that we are sure which ROCm version is used to refer to hcc/hip, etc. Going further multiple ROCm can be installed on a machine and when compiling any libs, we need to sure how the dependencies are met.

CMakeLists.txt
pruthvistony: Yes, Further by 3.1 or later, ROCm will always installed into /opt/rocm-<ver> folder and /opt/rocm should not used. So I updated this comments accordingly.

CMakeLists.txt
malcolmroberts: The user can just put the path to the ROCm cmake as the first element in the CMAKE_PREFIX_PATH; this guarantees that the user-specified path has highest priority.  

There is no reason to add another environment variable; things already work without this patch.  Moreover, not using standard cmake variables will almost certainly cause problems down the road.

Do you have a test environment with multiple rocm installs where you can try the branch that I provided?

library/src/CMakeLists.txt
pfultz2: The `CMAKE_SHARED_LINKER_FLAGS` variable should be changed by the toolchain and not in the CMakeLists.txt files.

pfultz2:
>I agree this implementation also works. Adding of ROCM_PATH is done so as to provide a uniform way of referring to all components of ROCm stack at any custom path.

We shouldn't add custom variables for this. The standard cmake variables(ie `CMAKE_PREFIX_PATH`, `CMAKE_INSTALL_PREFIX`, etc) should be used and it makes it much simpler.

CMakeLists.txt
pruthvistony: Before these changes, CMAKE_PREFIX_PATH was not used for searching all dependencies, the cmake file had hard coded /opt/rocm path at all place. So one of thought was to keep it in similar way so added ROCM_PATH (/opt/rocm-ver) to the search patch list. 
Will make the changes based on CMAKE_PREFIX_PATH and verify the running and get back.

library/src/CMakeLists.txt
pruthvistony: I will move it toolchain.

pruthvistony:
I will move the 

> 
> 
> > I agree this implementation also works. Adding of ROCM_PATH is done so as to provide a uniform way of referring to all components of ROCm stack at any custom path.
> 
> We shouldn't add custom variables for this. The standard cmake variables(ie `CMAKE_PREFIX_PATH`, `CMAKE_INSTALL_PREFIX`, etc) should be used and it makes it much simpler.

I will update the CMAKE_PREFIX_PATH and check the execution.

pruthvistony:
One information required about the ld.conf entries.
-> Are the ld.conf entries required when the package is executed/tested in CI machines.

For multiple ROCm installation to work properly, we are planning to remove all ld.conf entries in all the components. And all the dependencies should be found based on $ORIGIN in the RUNPATH.
Please let me know your opinion if it causes any problem for rocFFT.

pruthvistony:
I have updated the code based on the review comments.
Please check and let me know your comments and also please trigger the CI build, I dont know why it is not triggered.

CMakeLists.txt
```diff
@@ -96,10 +96,13 @@ elseif( CMAKE_CXX_COMPILER MATCHES ".*/hcc$" )
   message( STATUS "HCC compiler set; ROCm backend selected [ CXX=/opt/rocm/bin/hcc cmake ... ]" )
 endif( )
 
+# NOTE:  workaround until hcc & hip cmake modules fixes symlink logic in their config files; remove when fixed
+list( APPEND CMAKE_PREFIX_PATH ${ROCM_PATH} ${ROCM_PATH}/hcc ${ROCM_PATH}/hip )
```
pfultz2: This should be `list( APPEND CMAKE_PREFIX_PATH /opt/rocm /opt/rocm/hcc /opt/rocm/hip )`. This is for users building library. However, when building the packages for release it can be set to `-DCMAKE_PREFIX_PATH=/opt/rocm-<ver>;/opt/rocm-<ver>/hcc;/opt/rocm-<ver>/hip`, which it will search in `/opt/rocm-<ver>` first before searching in `/opt/rocm`.

PR #279, opened 2019-12-13
force exit on unknown plan

arghdos:
So fun fact here -- since the distributed version of the rocFFT library appears to be built w/o debug symbols, there is no type checking of the value you pass in as a hipfftPlan ([assert becomes a no-op](https://en.cppreference.com/w/cpp/error/assert) w/ NDEBUG defined).

If you are unlucky enough to fall into this trap, trying to execute the plan later will result in no kernels being launched!

Bit of an edge use-case, but this set me back for hours trying to port a Fortran code that used cuFFT to hipFFT (because fortran enums suck, and can't handle the hex-values in hipfftType!).

Best,
Nick

### DIFF
--- a/library/src/hipfft.cpp
+++ b/library/src/hipfft.cpp
@@ -317,7 +317,7 @@ hipfftResult hipfftMakePlan_internal(hipfftHandle            plan,
                                         dry_run));
         break;
     default:
-        assert(false);
+        return HIPFFT_PARSE_ERROR;
     }
 
     size_t tmpBufferSize = 0;


library/src/hipfft.cpp
```diff
@@ -317,7 +317,9 @@ hipfftResult hipfftMakePlan_internal(hipfftHandle            plan,
                                         dry_run));
         break;
     default:
-        assert(false);
+        std::stringstring ss;
+        ss << "hipfftType: " << type << " not recognized!");
+        throw std::runtime_exception(ss.str());
```
feizheng10: Throw exception from shared library is not a good practice for code portability.
Does "return HIPFFT_PARSE_ERROR" work for you?

arghdos: Yes, this totally works (and is more in keeping with the error checking status anyways)

arghdos: updated

PR #286, opened 2019-12-30
Add ekuznetsov139's patch for the 3.0 compilation issue.

### LINKED ISSUE #283: Build fails with compiler error on rocm 3.0
When compiling with hcc, via, for example,
```
cmake -DCMAKE_CXX_COMPILER=hcc ..
```
the compilation fails with some stack trace errors in '/opt/rocm/hcc/bin/clang-10'.

malcolmroberts:
resolves #283

### DIFF
--- a/library/src/device/generator/generator.pass.hpp
+++ b/library/src/device/generator/generator.pass.hpp
@@ -846,7 +846,8 @@ namespace StockhamGenerator
                             {
                                 passStr += rType + "wx, wy, rx, ry;\n\t\t";
                                 passStr += "wx = W.x; wy = W.y;\n\t\t";
-                                passStr += "rx = " + regRealIndex + "; ry = " + regImagIndex + ";\n\t\t";
+                                passStr += "rx = " + regRealIndex + "; ry = " + regImagIndex
+                                           + ";\n\t\t";
                                 if(fwd)
                                 {
                                     passStr += "TR = wx * rx - wy * ry;\n\t\t";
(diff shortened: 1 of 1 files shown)


malcolmroberts:
Failures seem to be related to docker issues and not htis patch.

library/src/device/generator/generator.pass.hpp
```diff
@@ -844,31 +844,19 @@ namespace StockhamGenerator
...
-                                    passStr += regRealIndex;
-                                    passStr += ") - (W.y * ";
-                                    passStr += regImagIndex;
-                                    passStr += ");\n\t\t";
-                                    passStr += "TI = (W.y * ";
-                                    passStr += regRealIndex;
-                                    passStr += ") + (W.x * ";
-                                    passStr += regImagIndex;
-                                    passStr += ");\n\t\t";
+                                    passStr += "TR = wx * rx - wy * ry;\n\t\t";
```
saadrahim: You are just rearranging expressions so the compiler bug is avoided?

malcolmroberts:
[figs.pdf](https://github.com/ROCmSoftwarePlatform/rocFFT/files/4010847/figs.pdf)
Looks like there's no effect on performance; so that's good!

library/src/device/generator/generator.pass.hpp
malcolmroberts: There's a local variable that serves as an explicit cast.

PR #294, opened 2020-03-17
Fix CMAKE_INSTALL_PREFIX in build_relocatable

zhang2amd:
CMAKE_INSTALL_PREFIX need to set to ${install_prefix} so that check_install_dir ${install_prefix} makes sense.

Summary of proposed changes:

- CMAKE_INSTALL_PREFIX set to $install_prefix for 'make install'
- CPACK_PACKAGING_INSTALL_PREFIX still set to $rocm_path for package install path
- 'make install' can be called by non-root user to test the build

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -428,7 +428,7 @@ if [[ "${build_cuda}" == false ]]; then
         cmake_client_options=" "
     fi
     if [[ "${build_relocatable}" == true ]]; then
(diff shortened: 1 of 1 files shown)


feizheng10:
@zhang2amd  Thanks for the fix.  I believe we need it ASAP.
Will merge it to branch master and hipclang-bkc-3. Do you need it for ROCm3.3 release?

zhang2amd:
@feizheng10 I currently only work on hipclang bkc3 and later bkc4, so I think master and hipclang-bkc-3 should be good.  ROCm 3.3 is not released yet so I don't know whether we need it. Maybe better merge in 3.3 also to save work later.

feizheng10:
@zhang2amd  master works as release candidate for ROCm3.3 release. So I guess it has to go into ROCm 3.3 release as well.

feizheng10:
Will merge into hipclang bkc3 and bkc4. 
Hold on to merge into master as per Subodh requested.

PR #302, opened 2020-06-30
Fix including of <cstddef>.

mhbliao:
- <cstddef> introduces C macros but, by itself, it's still a C++ header.

resolves SWDEV-241666


feizheng10:
PRed in internal repo.

PR #304, opened 2020-07-23
 use appropriate find_package calls for HIP

### LINKED ISSUE #303: [Build Error] [CMake] Cannot find hip
### What is the expected behavior
-  Including rocfft in a cmake project like find_package(rocfft) should be successful.

### What actually happens
-  When using find_package(rocfft), it fails since it cannot find hip, even though hip is installed.

```
CMake Error at /usr/share/cmake-3.18/Modules/CMakeFindDependencyMacro.cmake:47 (find_package):
  By not providing "Findhip.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "hip", but
  CMake did not find one.

  Could not find a package configuration file provided by "hip" with any of
  the following names:

    hipConfig.cmake
    hip-config.cmake

  Add the installation prefix of "hip" to CMAKE_PREFIX_PATH or set "hip_DIR"
  to a directory containing one of the above files.  If "hip" provides a
  separate development package or SDK, be sure it has been installed.
Call Stack (most recent call first):
  /opt/rocm/rocfft/lib/cmake/rocfft/rocfft-config.cmake:89 (find_dependency)
  cmake/public/LoadHIP.cmake:131 (find_package)
  cmake/public/LoadHIP.cmake:169 (find_package_and_print_version)
  cmake/Dependencies.cmake:995 (include)
  CMakeLists.txt:411 (include)
```

FindHIP.cmake is installed on the system but Findhip.cmake is not.

### How to reproduce
-  Call `find_package(rocfft)` in a cmake project.

### Environment
| Hardware | description |
|-----|-----|
| GPU | device string |
| CPU | device string |

| Software | version |
|-----|-----|
| ROCK | v3.5.1 |
| ROCR | v3.5.0 |
| HCC | N/A |
| Library | v0.0 | ???
| ROCFFT | v3.5.0 |

Downstream issue: https://github.com/pytorch/pytorch/issues/41886

acxz:
resolves #303 

Summary of proposed changes:
-  change find_package( hip ... calls to find_package( HIP ... to match the FindHIP.cmake file as installed by hip-base.
-  change export dep from hip to HIP

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,12 +115,12 @@ set(CMAKE_INSTALL_LIBDIR "lib" CACHE INTERNAL
 # Find HCC/HIP dependencies
 if( CMAKE_CXX_COMPILER MATCHES ".*/hcc$|.*/hipcc$" )
   message( STATUS "Building with ROCm tools" )
-  find_package( hip REQUIRED CONFIG PATHS )
+  find_package( HIP REQUIRED CONFIG PATHS )
 endif( )
 
 # Hip headers required of all clients; clients use hip to allocate
 # device memory
-find_package( hip REQUIRED CONFIG PATHS )
+find_package( HIP REQUIRED CONFIG PATHS )
 
 # CMake list of machine targets
 set( AMDGPU_TARGETS gfx803;gfx900;gfx906;gfx908 CACHE STRING
--- a/clients/CMakeLists.txt
+++ b/clients/CMakeLists.txt
@@ -69,8 +69,8 @@ endif( )
 # device memory.
 # NB: 2020-04-17: Using hip-clang compiled from source requires that
 # this not be added twice.
-if( NOT hip_FOUND)
-  find_package( hip REQUIRED CONFIG PATHS )
+if( NOT HIP_FOUND)
+  find_package( HIP REQUIRED CONFIG PATHS )
 endif()
 
 if( BUILD_CLIENTS_SAMPLES )
--- a/clients/samples/hipfft/CMakeLists.txt
+++ b/clients/samples/hipfft/CMakeLists.txt
@@ -22,7 +22,7 @@
 
 CMAKE_MINIMUM_REQUIRED(VERSION 3.0)
 
-find_package(hip)
+find_package(HIP)
 
 if( NOT TARGET rocfft )
   find_package(rocfft)
--- a/clients/samples/rocfft/CMakeLists.txt
+++ b/clients/samples/rocfft/CMakeLists.txt
@@ -22,7 +22,7 @@
 
 CMAKE_MINIMUM_REQUIRED(VERSION 3.0)
 
-find_package(hip)
+find_package(HIP)
 
 if( NOT TARGET rocfft )
   find_package(rocfft)
--- a/docs/samples/CMakeLists.txt
+++ b/docs/samples/CMakeLists.txt
@@ -37,7 +37,7 @@ foreach( exe ${samples} )
   target_include_directories( ${exe} PRIVATE ${rocfft_INCLUDE_DIR} )
   set_target_properties( ${exe} PROPERTIES CXX_STANDARD_REQUIRED ON)
   set_target_properties( ${exe} PROPERTIES CXX_STANDARD 14)
-  
+
   # prevent issue where __float128 is not supported
   set_target_properties( ${exe} PROPERTIES CXX_EXTENSIONS OFF)
 endforeach()
--- a/docs/samples/CMakeLists.txt
+++ b/docs/samples/CMakeLists.txt
@@ -25,7 +25,7 @@ CMAKE_MINIMUM_REQUIRED(VERSION 3.0)
 # project name
 PROJECT(rocfft_samples CXX)
 
-find_package(hip)
+find_package(HIP)
 find_package(rocfft)
 
 set (samples complex_1d complex_2d complex_3d real2complex_1d real2complex_2d real2complex_3d)
--- a/library/src/CMakeLists.txt
+++ b/library/src/CMakeLists.txt
@@ -111,7 +111,7 @@ endif( )
 # Installation
 
 rocm_install_targets(
-  TARGETS ${package_targets} 
+  TARGETS ${package_targets}
   INCLUDE
   ${CMAKE_SOURCE_DIR}/library/include
   ${CMAKE_BINARY_DIR}/include
--- a/library/src/CMakeLists.txt
+++ b/library/src/CMakeLists.txt
@@ -122,7 +122,7 @@ rocm_install_targets(
 rocm_export_targets(
   TARGETS roc::rocfft
   PREFIX rocfft
-  DEPENDS PACKAGE hip
+  DEPENDS PACKAGE HIP
   NAMESPACE roc::
   )
 


malcolmroberts:
Thanks for this, @acxz; I'll run this through CI and take a look.

acxz:
@malcolmroberts sounds good

pfultz2:
Lowercase "hip" is the correct name as the cmake config file is named [hip-config.cmake](https://github.com/ROCm-Developer-Tools/HIP/blob/master/hip-config.cmake.in). The problem in #303 is most likely due to something else. It may very likely be that the `CMAKE_PREFIX_PATH` is not set correctly as it as it mentions in the error:

```
Add the installation prefix of "hip" to CMAKE_PREFIX_PATH or set "hip_DIR"
  to a directory containing one of the above files.  If "hip" provides a
  separate development package or SDK, be sure it has been installed.
```

Which means you would need to set `CMAKE_PREFIX_PATH` to /opt/rocm/hip(but it needs to come before /opt/rocm because the symlinking in hip is broken).

TorreZuk:
ls /opt/rocm/hip/cmake/
FindHIP  FindHIP.cmake  FindROCR.cmake  FindROCT.cmake
But we use "HIP" for the find name convention and it is case sensitive.

acxz:
@pfultz2 my understanding is that `find_package(.)` and `DEPENDS PACKAGE ...` needs to be the same case as the `Find<..>.cmake` file, which is FindHIP.cmake in our case.

`find_package(.)` only throws warnings when using a different case name, but `DEPENDS PACKAGE ...` causes a failure when the `find_dependency(hip)` line is created in the exported cmake file for rocfft.

Manually renaming the `FindHIP.cmake` file to `Findhip.cmake` get's me farther along but still errors out.

The cmake config file being named lowercased `hip-..` does not matter.

acxz:
> But we use "HIP" for the find name convention and it is case sensitive.

@TorreZuk I am a bit confused by what you mean here?

TorreZuk:
My comment followed Paul's so was directed at him, @acxz I was agreeing with you.   Unless CMAKE_FIND_PACKAGE_PREFER_CONFIG is used we use the name  FindHIP.cmake (pasted the ls output) as the name so should use DEPENDS PACKAGE HIP.

pfultz2:
> my understanding is that find_package(.) and DEPENDS PACKAGE ... needs to be the same case as the Find<..>.cmake file, which is FindHIP.cmake in our case.

This should not be looking for FindHIP.cmake as this doesn't provide the hip::host and hip::device targets which are needed. It should be looking for the hip-config.cmake file.

If you are using FindHIP with `find_package(...)` then you will probably need to rename it to avoid this conflict.

Ideally, it would be best to have hip-config.cmake provide the functionality in FindHIP.cmake. This would avoid this naming conflict, and user's wouldn't need to copy FindHIP.cmake into their project. However, @mangupta strongly opposed this. So it seems the best option is to rename it when copying it into your project.

acxz:
> This should not be looking for FindHIP.cmake as this doesn't provide the hip::host and hip::device targets which are needed. It should be looking for the hip-config.cmake file.

Can you provide proof with this statement? I strongly believe `find_dependency` is looking for FindHIP.cmake and this is confirmed when I change the file name to Findhip.cmake and the find_dependecy call works.

> If you are using FindHIP with find_package(...) then you will probably need to rename it to avoid this conflict.

I am using the `FindHIP.cmake` installed by hip-base and adding the location to my CMAKE_MODULE_PATH. Expecting users to change a system level file is not ideal at all.

> Ideally, it would be best to have hip-config.cmake provide the functionality in FindHIP.cmake. 

Doesn't FindHIP.cmake call hip-config.cmake?

> This would avoid this naming conflict, and user's wouldn't need to copy FindHIP.cmake into their project.

What exactly is the naming conflict here? Copying FindHIP.cmake will still not fix the issue, since it needs to be Findhip.cmake to work with how the HIP dependency is currently exported. I am proposing we change our `hip` calls to `HIP` so that `Findhip.cmake` is not needed and upstream `FindHIP.cmake` can be used as is.

pfultz2:
> ls /opt/rocm/hip/cmake/
> FindHIP FindHIP.cmake FindROCR.cmake FindROCT.cmake
> But we use "HIP" for the find name convention and it is case sensitive.

Other teams have used that convention for the Find modules, but since this is only to support upstream projects you can name them whatever you want.(for example ROCR provides hsa-runtime-config.cmake so the find_package and Find module names do not need to match).

However, math libs and miopen need to find use hip-config.cmake because it uses the hip::host and hip::device targets to build with hip. This is why we call with the lowercase`find_package(hip)` to also make it clear we are looking for hip-config.cmake and not FindHIP.cmake.

acxz:
> However, math libs and miopen need to find use hip-config.cmake because it uses the hip::host and hip::device targets to build with hip. This is why we call with the lowercasefind_package(hip) to also make it clear we are looking for hip-config.cmake and not FindHIP.cmake.

hmm I see so the lower level issue is that FindHIP.cmake is not exporting all the targets in hip-config.cmake.

pfultz2:
> Can you provide proof with this statement? I strongly believe find_dependency is looking for FindHIP.cmake and this is confirmed when I change the file name to Findhip.cmake and the find_dependecy call works.

Well first, there is most likely a bug in cmake when one uses an uppercase Find module with a lowercase package config file. However, these cmake files are definitely looking for hip-config.cmake as they need the hip::host and hip::device function. 

> I am using the FindHIP.cmake installed by hip-base and adding the location to my CMAKE_MODULE_PATH. Expecting users to change a system level file is not ideal at all.

Find modules are to support upstream projects not downstream projects. That is the project has a Find module in its own repo and uses that module, so there is no need to change a system level file.

Find modules are not meant to support downstream projects. Trying to force it to support downstream projects through `CMAKE_MODULE_PATH` is problematic as it breaks cmake's standard searching mechanism for dependencies(ie `CMAKE_PREFIX_PATH` or `<PackageName>_DIR`.

> Doesn't FindHIP.cmake invoke the call to hip-config.cmake in the user cmake project?

No it doesn't.

> What exactly is the naming conflict here? Copying FindHIP.cmake will still not fix the issue, since it needs to be Findhip.cmake to work with how the HIP dependency is currently exported. I am proposing we change our hip calls to HIP so that Findhip.cmake is not needed and upstream FindHIP.cmake can be used as is.

For projects that need to use special cmake functions like `hip_add_library`, they should copy FindHIP.cmake into their project and rename it to something like FindHIP2.cmake and then call `find_package(HIP2)`. This way the projects can call `find_dependency(hip)` without an ambiguity.

pfultz2:
> hmm I see so the lower level issue is that FindHIP.cmake is not exporting all the targets in hip-config.cmake.

Actually its the reverse. hip-config.cmake needs to include the functionality in FindHIP.cmake. hip-config.cmake is what is available to support downstream `find_package(hip)`.

acxz:
Gotcha thanks for explaining @pfultz2 

Well then this seems to be a user issue then. When calling cmake I (as well as downstream projects like [pytorch](https://github.com/pytorch/pytorch/issues/41886) should ensure that `hip-config.cmake` is used over `FindHIP.cmake`. Is that correct?

pfultz2:
> Well then this seems to be a user issue then. When calling cmake I (as well as downstream projects like pytorch should ensure that hip-config.cmake is used over FindHIP.cmake. Is that correct?

Yes, but some downstream projects(like pytorch) use the `hip_add_library` cmake functions which currently hip-config.cmake doesn't provide. So we need to update hip-config.cmake to provide this same functionality.

acxz:
I see, then an issue and corresponding PR should be made over at the HIP repo about this.

saadrahim:
Should this issue be close?

malcolmroberts:
We have separately added these changes; I am going to close this PR.

acxz:
Thanks @malcolmroberts !

PR #305, opened 2020-08-19
fixes some bugs with incorrect calculation results

hoptony:
resolves #___

Summary of proposed changes:
-  1.fixes incorrect calculation results of rocfft when the stride is not equal to 1;(modified 
      generator.kernel.hpp, transpose.h, plan.cpp )
-  2.fixes the incorrect calculation results of some large general length data of non radix-2, radix-3 
      and radix-5(eg:length=40353607,47045881);(modified plan.cpp  )
-  3.fixes the 2D-FFT's incorrect calculation results of some large 
      length(eg:lenx=32*1048576,leny=1); (modified powX.cpp)
-  4.fixes the 3D-FFT's incorrect calculation results of some large 
      length(eg:lenx=32*1048576,leny=1,lenz=1) (modified powX.cpp);

### DIFF
--- a/library/src/device/generator/generator.kernel.hpp
+++ b/library/src/device/generator/generator.kernel.hpp
@@ -307,25 +307,33 @@ namespace StockhamGenerator
             loop += stride_name1 + "[2] + (counter_mod % " + sub_string + ")*"
                     + std::to_string(blockWidth);
             if(blockComputeType == BCT_R2C) // only for input
-                loop += "*lengths[0]";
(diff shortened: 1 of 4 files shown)


malcolmroberts:
Thanks for this!  We will take a look at this and make sure it passes CI.

feizheng10:
Thanks for great contribution on rocFFT!  We will evaluate one by one for both correctness and performance.
Still, we really suggest to separate these into a few PRs.

PR #318, opened 2021-06-17
Fix old sbcc and make old/new sbcc co-exist

solaslin:
back port from latest branch: switch len 100 to old SBCC

### DIFF
--- a/library/src/device/generator/generator.kernel.hpp
+++ b/library/src/device/generator/generator.kernel.hpp
@@ -1158,7 +1158,7 @@ namespace StockhamGenerator
             // Function signature
             if(NeedsLargeTwiddles())
             {
-                str += "template <typename T, StrideBin sb, bool TwdLarge, CallbackType cbtype, "
+                str += "template <typename T, StrideBin sb, CallbackType cbtype, bool TwdLarge, "
                        "size_t LTBase="
                        + std::to_string(LTWD_BASE_DEFAULT) + ">\n";
             }
--- a/library/src/plan.cpp
+++ b/library/src/plan.cpp
@@ -1007,21 +1007,41 @@ bool TreeNode::use_CS_2D_SINGLE()
 
 bool TreeNode::use_CS_2D_RC()
 {
-    //   For CS_2D_RC, we are reusing SBCC kernel for 1D middle size. The
-    //   current implementation of 1D SBCC supports only 64, 128, and 256.
-    //   However, technically no LDS limitation along the fast dimension
-    //   on upper bound for 2D SBCC cases, and even should not limit to pow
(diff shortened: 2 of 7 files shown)


evetsso (approved):

memmett (approved):

feizheng10 (approved):

library/src/device/generator/generator.kernel.hpp
```diff
@@ -1158,7 +1158,7 @@ namespace StockhamGenerator
             // Function signature
             if(NeedsLargeTwiddles())
             {
-                str += "template <typename T, StrideBin sb, bool TwdLarge, CallbackType cbtype, "
+                str += "template <typename T, StrideBin sb, CallbackType cbtype, bool TwdLarge, "
```
malcolmroberts, line 1161: This was switched for some reason?

library/src/plan.cpp
```diff
@@ -1007,21 +1007,41 @@ bool TreeNode::use_CS_2D_SINGLE()
...
-    //   However, technically no LDS limitation along the fast dimension
-    //   on upper bound for 2D SBCC cases, and even should not limit to pow
-    //   of 2.
-
-    // FIXME: use all SBCC kernels instead after we fix the bugs in buffer assignment
-    // std::set<int> sbcc_support = {50, 64, 81, 100, 128, 200, 256};
-    std::set<int> sbcc_support = {64, 128, 256};
-    if((sbcc_support.find(length[1]) != sbcc_support.end()) && (length[0] >= 64))
+    try
+    {
```
malcolmroberts, line 1011: This looks like we should expect the try block to fail under normal circumstances.  Using exceptions for normal flow seems like a bad idea; exceptions should be for exceptional circumstances.

library/src/device/generator/generator.kernel.hpp
evetsso: I put cbtype before TwdLarge when callbacks were merged.  But that was after sbcc was moved to the new generator, so the old generator wasn't updated.  This change is updating the old generator to match what the new generator is now doing.

library/src/plan.cpp
evetsso: Rearranging that should be the subject of a different PR, though.

malcolmroberts:
Is there any impact on other tests?  Please run the performance suite on MI50 and MI100.

memmett:
It's 5am in Taiwan.  I can run on MI100...

malcolmroberts:
MI50 looks good.
[figs.pdf](https://github.com/ROCmSoftwarePlatform/rocFFT/files/6678490/figs.pdf)

PR #334, opened 2021-10-29
Updating to Googletest 1.11.0

pavahora:
resolves #___

Summary of proposed changes:
-  
-  
-

### DIFF
--- a/clients/CMakeLists.txt
+++ b/clients/CMakeLists.txt
@@ -87,7 +87,7 @@ if( BUILD_CLIENTS_TESTS  OR BUILD_CLIENTS_SELFTEST )
 
     ExternalProject_Add( gtest
       URL https://github.com/google/googletest/archive/release-1.11.0.tar.gz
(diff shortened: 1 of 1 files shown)


clients/CMakeLists.txt
```diff
@@ -86,7 +86,7 @@ if( BUILD_CLIENTS_TESTS  OR BUILD_CLIENTS_SELFTEST )
       )
 
     ExternalProject_Add( gtest
-      URL https://github.com/google/googletest/archive/release-1.10.0.tar.gz
+      URL https://github.com/google/googletest/archive/release-1.11.0.tar.gz
       URL_HASH SHA256=9dc9157a9a1551ec7a7e43daea9a694a0bb5fb8bec81235d8a1e6ef64c716dcb
```
evetsso, line 90: This hash also needs to be updated or else the download verification will fail.  Looks like the correct hash should be `b4870bf121ff7795ba20d20bcdd8627b8e088f2d1dab299a031c1034eddc93d5`.

pavahora: Updated new hash

PR #348, opened 2022-03-02
add length 512 SBCC+SBRC, enable for 1D 2^17 and 2^18

evetsso:
Cherry-pick for 5.1, as the release branch does not contain this work and it's necessary for 5.1.


evetsso:
I've manually confirmed that we're right around where we expect with this fix on MI200 - ~306ms.

PR #355, opened 2022-04-22
Don't use sbcc 243 on 243^3

solaslin:
resolves #SWDEV-333618


solaslin:
We prefer merging into release/rel-rocm-5.2 than master, since this is a workaround for 5.2 only, 5.3 will not have this issue. So put it as draft for now.

PR #379, opened 2022-10-12
Add gfx1101 to default AMDGPU_TARGETS

cheyennegoh:
Summary of proposed changes:
-  Adds gfx1101 (Navi 32) to default AMDGPU_TARGETS


TorreZuk (commented):
@cheyennegoh gfx1101 is only being enabled in develop branch I believe.  You are targeting rocm-rel-5.4 so just wanted to make sure this is intended.

cheyennegoh:
> @cheyennegoh gfx1101 is only being enabled in develop branch I believe. You are targeting rocm-rel-5.4 so just wanted to make sure this is intended.

@TorreZuk Thanks for bringing this to our attention. We'll close this PR.

PR #384, opened 2022-10-29
reduce aot build time

evetsso:


.jenkins/common.groovy
```diff
@@ -75,7 +75,7 @@ def runTestCommand (platform, project, boolean debug=false)
     def command = """#!/usr/bin/env bash
                 set -x
                 cd ${project.paths.project_build_prefix}/build/${directory}/clients/staging
-                ROCM_PATH=/opt/rocm GTEST_LISTENER=NO_PASS_LINE_IN_LOG ./${testBinaryName} --gtest_color=yes --R 80 --gtest_filter='-rocfft_UnitTest.repo_twiddle'
+                ROCM_PATH=/opt/rocm GTEST_LISTENER=NO_PASS_LINE_IN_LOG ./${testBinaryName} --precompile --gtest_color=yes --R 80 --gtest_filter='-rocfft_UnitTest.repo_twiddle'
```
feizheng10, line 78: Not related to this PR, just a thought,  no "/opt/rocm" even in CI.

feizheng10 (approved):
LG

.jenkins/common.groovy
evetsso: Hmm.. yeah that shouldn't really be necessary, or at least if it is, something outside of this script should be setting it.

clients/tests/gtest_main.cpp
```diff
@@ -85,6 +90,70 @@ static size_t get_system_memory_GiB()
...
+            {
+                name.erase(0, 8);
+                tokens.emplace_back(std::move(name));
+            }
+        }
+    }
+
+    std::random_device dev;
+    std::mt19937       dist(dev());
+    std::shuffle(tokens.begin(), tokens.end(), dist);
```
malcolmroberts, line 122: This is to even out the load between the processes?

clients/tests/accuracy_test_adhoc.cpp
```diff
@@ -39,15 +39,6 @@ std::vector<std::vector<size_t>> adhoc_sizes = {
     // Failure with build_CS_3D_BLOCK_RC
     {680, 128, 128},
 
-    // Large 1D primes that fall above the block threshold (length 262144).
```
malcolmroberts, line 42: Why were these test cases removed?

malcolmroberts (approved):

clients/tests/accuracy_test_adhoc.cpp
evetsso: They were only accidentally added (per @eng-flavio-teixeira ) and they cause buffer assignment to explode with combinations, using up all your host memory (assuming the test actually decides to try building the plan).  Precompilation does not check if the problem will fit, so it hits this problem.

Limiting the explosion would be helpful but is a separate problem, and once that's addressed we can consider re-adding this case.  But it won't fit on any GPUs we have currently.

eng-flavio-teixeira: The accidental part of this was not realizing that by adding the entry {196597, 25165813} to adhoc_sizes, it would run the 2D problem 196597 x 25165813. The intention **was to only test the 1D problems 196597 and 25165813**. 

Could we just add the entries separately {196597} and {25165813} to adhoc_sizes, so that it tests only the 1D cases?

eng-flavio-teixeira (approved):

clients/tests/accuracy_test_adhoc.cpp
evetsso: Ah ok, we'll need to fix that in develop too then.  I misunderstood you earlier.

clients/tests/gtest_main.cpp
evetsso: The shuffle is to try to make sure many similar test cases aren't done in parallel, which might cause us to compile the same kernel at the same time in multiple threads.  Currently the RTC code doesn't try to prevent that.  I hope to make that change eventually though, and then this shuffle could go away.

amd-aakash:
@malcolmroberts / @feizheng10 / @eng-flavio-teixeira - This PR has been approved, can you please help merge this into the release branch.

PR #386, opened 2022-11-07
Correct hip path for new directory layout

raramakr:
SWDEV-345870 - Use actual hip installation path /opt/rocm-ver rather than using backward compatibility path /opt/rocm-ver/hip

resolves #___

Summary of proposed changes:
-  
-  
-


raramakr:
Raised the PR in rocfft-internal branch

PR #391, opened 2023-01-17
HOTFIX: Remove extra header location from install call

lawruble13:
resolves SWDEV-362068

Summary of proposed changes:

- Remove source directory from rocm_install_targets call to prevent installation of rocfft.h in an unintended location


lawruble13:
Hotfix approved by Makarand.

cgmb:
If there's going to be a new version of rocfft in ROCm 5.4.3, when do you plan to update the library version number and update the changelog?

PR #398, opened 2023-02-12
Add missing type definitions

littlewu2508:
This fixes compilation errors

### DIFF
--- a/library/src/device/kernels/memory_gfx.h
+++ b/library/src/device/kernels/memory_gfx.h
@@ -68,6 +68,8 @@ struct CacheOperation
 
 using float16_t = _Float16;
 using float32_t = float;
+using uint32_t = unsigned int;
+using int32_t = int;
 
 template <typename T, int N>
 struct NativeVector
--- a/shared/array_validator.h
+++ b/shared/array_validator.h
@@ -24,8 +24,8 @@
 #include <vector>
 
 // Checks whether the array with given length and stride has multi-index collisions.
-bool array_valid(const std::vector<size_t>& length,
-                 const std::vector<size_t>& stride,
+bool array_valid(const std::vector<std::size_t>& length,
+                 const std::vector<std::size_t>& stride,
                  const int                  verbose = 0);
 
 #endif


littlewu2508:
The compiler does not know about types like uint32_t, a simple program like
```c++
int main() {
        uint32_t a;
        return 0;
}
```
this will yield 
```
main.C: In function ‘int main()’:
main.C:2:9: error: ‘uint32_t’ was not declared in this scope
    2 |         uint32_t a;
      |         ^~~~~~~~
main.C:1:1: note: ‘uint32_t’ is defined in header ‘<cstdint>’; did you forget to ‘#include <cstdint>’?
```

They are defined in `stdint.h`, so probably this header should include <cstdint>

cgmb:
This may be a difference between amdclang and upstream clang. The AMD Clang will include stdint.h by default when compiling with `-x hip`, but Clang 15.0.7 does not seem to do so.

evetsso:
@littlewu2508 Can you confirm if you were using AMD Clang or upstream for this build?

Including `cstdint` sounds like a better way forward here.

littlewu2508:
> Can you confirm if you were using AMD Clang or upstream for this build?

I'm using upstream clang

evetsso:
@littlewu2508 I'm unable to reproduce a compilation error building the rocFFT `rocm-5.4.3` tag (commit 5687cd9dc9fa27e92f9997e877486b4798eab4b2) using upstream clang 15.0.7 (as provided by Arch Linux).

Can you provide more detail on exactly what errors you saw?  Which compilation unit failed?  And are you able to report the command line passed to the compiler when this failure happened?

cgmb:
@littlewu2508, I'm trying to make that patch unnecessary. [There is support for an FHS-compliant path for the HIP version file in clang trunk](https://github.com/llvm/llvm-project/commit/082593ff7aff68060bd66dccfa43493d07d9c255), but has not been backported to LLVM 15, and it is limited to Debian or Red Hat. The distro guards on that functionality are a problem for Ubuntu 23.04. Maybe they're an issue for Gentoo, too?

littlewu2508:
> @littlewu2508, I'm trying to make that patch unnecessary. [There is support for an FHS-compliant path for the HIP version file in clang trunk](https://github.com/llvm/llvm-project/commit/082593ff7aff68060bd66dccfa43493d07d9c255), but has not been backported to LLVM 15 and it is limited to Debian or Red Hat. The distro guards on that functionality are a problem for Ubuntu 23.04. Maybe they're an issue for Gentoo, too?

Exactly. The HIP detection is done either by detecting `.hipVersion` or passing `--hip-version` to clang. Using [There is support for an FHS-compliant path for the HIP version file in clang trunk](https://github.com/llvm/llvm-project/commit/082593ff7aff68060bd66dccfa43493d07d9c255) or https://github.com/gentoo/gentoo/commit/bc4899922bfb97416027d18a049f00bc8663fe86 (`--hip-version` patch adopted from debian) works and make this PR unnecessary.

PR #432, opened 2023-07-26
hotfix: fixed static regEx (5.6.1)

solaslin:
resolves #412803  https://ontrack-internal.amd.com/browse/SWDEV-412803

- Fixed as per the rocfft-patch.diff
- Same patch would be applied on 5.7 as well (in another PR)

PR #433, opened 2023-07-26
hotfix: fixed static regEx (5.7)

solaslin:
Same defect as #432 
Cherry-picking to 5.7 branch

### DIFF
--- a/library/src/include/option_util.h
+++ b/library/src/include/option_util.h
@@ -35,10 +35,8 @@
 #include <vector>
 
 // Regular expression for token delimiters (whitespace and commas)
-static const std::regex program_options_regex{"[, \\f\\n\\r\\t\\v]+",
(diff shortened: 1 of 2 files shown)


evetsso:
I believe this should target release-staging/rocm-rel-5.7, not release/rocm-rel-5.7?

solaslin:
> I believe this should target release-staging/rocm-rel-5.7, not release/rocm-rel-5.7?

Thanks, I'll change it. Could you please point me out when to target release-staging/ and when to release/ directly ? 
I guess in this PR, I target this to release-staging/ first, and see if it passes the staging, and then another PR to release/ , is that right? So how about the 5.6 one ? I see the release-staging/ hasn't been updated for a very long time.

evetsso (approved):

PR #434, opened 2023-07-26
cmake: install rtc helper to versioned subdirectory

evetsso:
Since that's where rocFFT will look for it.


evetsso:
Wrong branch.

PR #438, opened 2023-08-02
Silence python invalid escape warning

trixirt:
On Fedora/rawhide with Python 3.12.0b4
There is this warning
rocFFT/library/src/device/kernel-generator-embed-cpp.py:63: SyntaxWarning: invalid escape sequence '\s'
  include_regex = re.compile('''^\s*#include''')

Change to a raw string


evetsso (approved):
Thanks for the fix!

PR #443, opened 2023-11-22
Add CMake Option To Disable Kernel Cache

tflink:
The kernel cache takes a long time to build and it poses some challenges when packaging rocFFT for linux distributions. It would be nice if there was an option to disable the cache db building entirely.

Summary of proposed changes:
-  adds `ROCFFT_KERNEL_CACHE_ENABLE` option to cmake (default to `ON` which preserves existing default behavior) that disables kernel cache creation and install if set to `OFF`

### DIFF
--- a/library/src/CMakeLists.txt
+++ b/library/src/CMakeLists.txt
@@ -497,7 +497,11 @@ if( ROCFFT_BUILD_OFFLINE_TUNER )
 endif()
 
 # compile kernels into the cache file we ship
-#
+
+# While useful in most situations, building the kernel cache takes a long time
+# enable a configure-time option to skip kernel cache building
+option( ROCFFT_KERNEL_CACHE_ENABLE "Enable building rocFFT kernel cache" ON)
+
 # cache file should go next to the shared object - on Windows this
 # would be the DLL, not the import library.
 if( WIN32 )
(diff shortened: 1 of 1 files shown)


evetsso (approved):
LGTM.  Thanks for the contribution!

cgmb:
> The kernel cache takes a long time to build and it poses some challenges when packaging rocFFT for linux distributions.

@tflink, could you elaborate on this? Would it help if there were one cache per architecture, so users can install only kernels they need? Or is the problem related to something else?

evetsso:
There isn't really any getting around the fact that building a lot of kernels takes a lot of time.  Distros like Fedora are building kernels for more architectures than rocFFT's CMakeLists.txt asks for by default, so it's not surprising to me that this cost is painful.

Currently, the way rocFFT is written is we generate code for a kernel, then invoke hipRTC N times to build for N architectures.  We could reorganize rocFFT so that hipRTC is told about all N architectures up front, so the compilation only needs to happen once, and then code for N architectures is emitted together.  There's some design work we'd need to do to make this happen but I've got it on the TODO list.  Plus we'd need a POC to determine how much of a win this will end up being.

But for the time being, having an easy switch to turn the whole thing off is reasonable IMO.

cgmb:
> Currently, the way rocFFT is written is we generate code for a kernel, then invoke hipRTC N times to build for N architectures. We could reorganize rocFFT so that hipRTC is told about all N architectures up front, so the compilation only needs to happen once, and then code for N architectures is emitted together. There's some design work we'd need to do to make this happen but I've got it on the TODO list. Plus we'd need a POC to determine how much of a win this will end up being.

Where does the speedup come from in that case? I thought that when you invoke clang with multiple architectures, it basically just calls itself multiple times and bundles the results together.

> But for the time being, having an easy switch to turn the whole thing off is reasonable IMO.

Agreed.

evetsso:
> > Currently, the way rocFFT is written is we generate code for a kernel, then invoke hipRTC N times to build for N architectures. We could reorganize rocFFT so that hipRTC is told about all N architectures up front, so the compilation only needs to happen once, and then code for N architectures is emitted together.
> 
> Where does the speedup come from in that case? I thought that when you invoke clang with multiple architectures, it basically just calls itself multiple times and bundles the results together.

This is the concept I'd need to prove.  I thought the parsing, translation to IR, and (some?) optimization passes would only happen once, and code generation would happen N times from the optimized IR.  This might not turn out to be true, or might not be as true as I'm hoping.

tflink:
> > The kernel cache takes a long time to build and it poses some challenges when packaging rocFFT for linux distributions.
> 
> @tflink, could you elaborate on this? Would it help if there were one cache per architecture, so users can install only kernels they need? Or is the problem related to something else?

Before I answer, I want to clarify that I don't mean any of this as criticism,  just observations from the POV of someone trying to interface two different things with overlapping but not identical concerns. If I've misunderstood something, please correct me.

For the packaging bit, I'm still not 100% sure where the kernel db should go if we get more pedantic. It's not a library, so it doesn't belong in `/usr/lib`, it's not an executable and doesn't belong in `/usr/bin`. The closest match I can think of is `/usr/libexec` but I didn't see a way to move the kernel db without patching the code and that's more work than I want to deal with at the moment. AFAIK, rocFFT isn't the only rocm package that has this issue (I believe rocBLAS does something similar) so I plan to revisit the "where should these things live?" question once we have everything packaged and working in Fedora.

I was under the impression that debian disabled the kernel cache for their rocFFT package for the same reason - did you all find a solution?

With regards to the time thing, that's just a convenience bit as far as I'm concerned. Build time doesn't matter a ton for the official packages; even with the kernel cache enabled, rocFFT is far from the longest build time for Fedora packages. It's just that if I'm building locally, I don't always want to wait for the kernel caches to build.

Getting back to the "why?" part, the primary reason for the PR was that I don't know where to put the kernel cache  for the Fedora package and will be disabling it for the near future, at least. I don't really want to maintain out-of-tree patch if I don't have to and Fedora packaging guidelines heavily discourage carrying patches which aren't at least proposed upstream. Those two reasons were enough for me to at least propose it upstream to see if there was interest. I also figured that I wasn't the only one who didn't want to wait for the kernel cache to build every time so that part was an added bonus, in a way :)

cgmb:
> This is the concept I'd need to prove. I thought the parsing, translation to IR, and (some?) optimization passes would only happen once, and code generation would happen N times from the optimized IR. This might not turn out to be true, or might not be as true as I'm hoping.

I don't think it can because [clang defines different macros](https://clang.llvm.org/docs/AMDGPUSupport.html#predefined-macros) depending on what architecture is being built (e.g. `__gfx1030__` is only defined in the pass for gfx1030 and `__gfx906__` is only defined for the pass for gfx906).

cgmb:
> I was under the impression that debian disabled the kernel cache for their rocFFT package for the same reason - did you all find a solution?

No. Not yet.

evetsso:
> For the packaging bit, I'm still not 100% sure where the kernel db should go if we get more pedantic. It's not a library, so it doesn't belong in `/usr/lib`, it's not an executable and doesn't belong in `/usr/bin`. The closest match I can think of is `/usr/libexec` but I didn't see a way to move the kernel db without patching the code and that's more work than I want to deal with at the moment.

I think pretty soon we'll need to split the rocFFT package anyway, and create some kind of `rocfft-data` package which has the kernels (if one chooses to build them).  At that time it would be reasonable to move the cache to libexec or wherever else makes sense.  If it's a location that's easily chosen at build time then that'll probably make everyone happy.

So as long as you don't mind waiting a bit, I don't think this is something you'll need to think about too hard.

tflink:
Good to know that it's something that's likely to be worked on, thanks.

There are plenty of other things on my TODO list (like getting some testing infrastructure in place) so I doubt that waiting will be an issue :)

PR #446, opened 2023-12-08
device: uint32_t -> __hip_uint32_t, int32_t -> __hip_int32_t

evetsso:
The undecorated type names are deprecated in hipRTC.


feizheng10 (approved):
Does this block run our code on NV cards?

af-ayala (approved):

evetsso:
> Does this block run our code on NV cards?

It's certainly not helping.  The typedefs are being removed because they're not supposed to be there, and *that* was breaking some users that were defining their own equivalents.  So with this change, we're more compatible with NV.

eng-flavio-teixeira (approved):

malcolmroberts (approved):

evetsso:
Type removal in hipRTC didn't end up happening in 6.0.  Closing.

PR #455, opened 2024-01-25
Add -mcumode arg while compiling with hiprtc

cjatin:
This resolves some performance issues seen with Navi GPUs

Summary of proposed changes:
-  This is required due to recent change in hiprtc which removes the arg `-mcumode` added by default
- https://github.com/ROCm/clr/commit/49369f08519be2c8643736c5611c9cd77ce47b93
- This only affects Navi GPUs since gfx9 only support CU mode


cjatin:
closing this

PR #471, opened 2024-03-15
rtc: use gridDim.x for everything in real-complex kernels 

af-ayala:
Fixes launch errors on very large odd-length real-complex transforms, since
gridDim.y and gridDim.z are limited to 64k.


feizheng10 (approved):
LG. Please double check the accuracy tests and perf test results.

PR #476, opened 2024-04-10
Documentation: Remove faulty reference

MKKnorr:
The (develop) documentation currently faultily links to rocblas, as :ref:`How-To` is currently just a place holder, and intersphinx then starts looking for the reference in other linked projects, even if the reference isn't scoped properly.

The placeholder references are removed for now, and to avoid that problem occuring again, the `external_projects` have been set to an empty list. If rocFFTs documentation needs to link to other projects, they have to be added to that list (see [rocm-docs-core for reference](https://rocm.docs.amd.com/projects/rocm-docs-core/en/latest/developer_guide/projects_yaml.html))

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -15,20 +15,22 @@ The code is open and hosted here: https://github.com/ROCmSoftwarePlatform/rocFFT
 
 The rocFFT documentation is structured as follows:
 
-.. card:: Conceptual
+.. grid:: 2
 
-  * :ref:`what-is-rocfft`
+    .. grid-item-card:: Conceptual
 
-.. card:: How To
+      * :ref:`what-is-rocfft`
 
-  * :ref:`working-with-rocfft`
-  * :ref:`load-store-callbacks`
-  * :ref:`runtime-compilation`
+    .. grid-item-card:: How To
 
-.. card:: API Reference
+      * :ref:`working-with-rocfft`
+      * :ref:`load-store-callbacks`
+      * :ref:`runtime-compilation`
 
-  * :ref:`api-usage`
-  * :ref:`api-reference`
+    .. grid-item-card:: API Reference
+
+      * :ref:`api-usage`
+      * :ref:`api-reference`
 
 To contribute to the documentation refer to `Contributing to ROCm  <https://rocm.docs.amd.com/en/latest/contribute/contributing.html>`_.
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -15,17 +15,17 @@ The code is open and hosted here: https://github.com/ROCmSoftwarePlatform/rocFFT
 
 The rocFFT documentation is structured as follows:
 
-.. card:: :ref:`Conceptual`
+.. card:: Conceptual
 
   * :ref:`what-is-rocfft`
 
-.. card:: :ref:`How-To`
+.. card:: How To
 
   * :ref:`working-with-rocfft`
   * :ref:`load-store-callbacks`
   * :ref:`runtime-compilation`
 
-.. card:: :ref:`API Reference`
+.. card:: API Reference
 
   * :ref:`api-usage`
   * :ref:`api-reference`
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -24,6 +24,7 @@ docs_core.run_doxygen(doxygen_root="doxygen", doxygen_path="doxygen/xml")
 docs_core.setup()
 
 external_projects_current_project = "rocfft"
+external_projects = []
 
 for sphinx_var in ROCmDocs.SPHINX_VARS:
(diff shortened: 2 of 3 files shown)


randyh62 (approved):

docs/index.rst
randyh62, line 1: The layout of the tiles or cards in the index.rst file has also changed to implement a multi-column look with a gutter for space between tiles. You can see the new structure at: https://raw.githubusercontent.com/ROCm/rocSOLVER/develop/docs/index.rst

Rmalavally (approved):
@MKKnorr, how does this change impact content that has already been refactored, if it does? @randyh62 We must update our internal guidelines to ensure all writers are aware.

docs/index.rst
MKKnorr: This is not within the scope of this PR, but since it is a very small change I included it anyways. I also added the files from the `how-to` directory to the table of contents.

Btw, since you seem to have restructured this, if I've seen it right, do you have any input as for why the files from the `design/` directory are not included anymore? That seems to be quite valuable information. Also it is emitting warnings right now during the build, that these files aren't included anywhere.

randyh62: The design files were removed because they were written as engineering proposals rather than documentation of features. In the discussions at the the time it was agreed they could be referenced from the README files for feedback from community if desired.

Rmalavally:
> > We must update our internal guidelines to ensure all writers are aware.
> 
> That is a good idea. Can you tell me where those internal guidelines are? I'd be happy to contribute, and it also wouldn't hurt being aware of them myself.
> 
> > how does this change impact content that has already been refactored, if it does?
> 
> This does not affect content that has already been refactored, this is a fix for an oversight that was introduced during the refactoring.
> 
> To be specific: when looking at the [rendered version of the restructured documentation](https://rocm.docs.amd.com/projects/rocFFT/en/develop/) you can see that the second card is titled "What is rocBLAS?" and actually links to rocBLAS, instead of rocFFT.
> 
> This was due to the fact, that the references in the titles have not been defined ([see :ref:`Conceptual`, :ref:`How-To` and :ref:`API Reference`](https://raw.githubusercontent.com/ROCm/rocFFT/61a53cdb7ffe3212cf39b1668b9a0d5355df691e/docs/index.rst)
> 
> Usually, this causes a warning to be emitted when building the documentation, the reference will be ignored, and only the plain text will be printed there, as can be seen in the other two cards.
> 
> However in this specific circumstance, rocBLAS happened to define a reference for :ref:`How-To`, and since the rocm documentation by default includes the references from all other rocm repositories (see the list in [rocm-docs-core's projects.yaml](https://github.com/ROCm/rocm-docs-core/blob/develop/src/rocm_docs/data/projects.yaml)), this has been picked up, and so the title actually linked to rocBLAS' documentation.
> 
> There are two ways to circumvent this:
> 
> 1. Not including all the other repositories' references, when it is not actually needed
> 
> In the documentation's config.py, which is located in each documentation's root directory, the external projects that should be linked to can be set by including the names of the projects in the `external_projects` list. To not link to any other projects, which rocm-docs-core does by default, the variables can be set to the empty list, as has been done in this PR. This is explained in more detail [on rocm-docs-core's documentation](https://rocm.docs.amd.com/projects/rocm-docs-core/en/latest/developer_guide/projects_yaml.html#explicitly-list-external-projects)
> 
> 2. prefixing all references with a repository specific tag (e.g. rocfft in this case)
> 
> This might not be feasible, and is easy to overlook. Especially when considering, that this kind of mix-up does not give a warning when building the documentation, and in this case even the faulty reference in the rendered version has not been caught during review.

@MKKnorr Thanks for the detailed explanation. i created this internal page to capture observations and guidelines for refactoring, so writers are aligned for consistency,

ttps://confluence.amd.com/display/MLSE/Refactoring+Documentation+Guidelines

docs/index.rst
MKKnorr: Okay, I think that's fine for now. I still think, that just not including them is not a good way of handling this. It will generate warnings when building the documentation, which makes it harder to spot the warnings that apply to the new changes one is actually working on.

yhuiYH (approved):
Looks good. thx for the fix.

lpaoletti:
Took a look and noticed the following:
image

image

image

feizheng10 (approved):

MKKnorr:
@lpaoletti 
> Took a look and noticed the following:

I agree, that there are some inconsistencies, and it would be nice to have had them fixed up when this documentation was restructured, but this PR is just supposed to fix the faulty references. Please open another issue for the ROCm documentation team for this

MKKnorr:
This PR is being superseded by https://github.com/ROCm/rocFFT-internal/pull/1277 in the internal repo

PR #488, opened 2024-06-21
Support gfx1200 and gfx1201

malcolmroberts:


malcolmroberts:
Moved to internal for the normal flow.

PR #509, opened 2024-08-23
doc: add some content about distributed transforms

evetsso:


feizheng10 (approved):

af-ayala (approved):

docs/how-to/working-with-rocfft.rst
```diff
@@ -344,7 +347,7 @@ The workflow for fields is as follows:
 #. Create the plan by calling :cpp:func:`rocfft_plan_create`.  Pass the plan
    description that has already been allocated.
 
-#. Execute the plan by calling :cpp:func:`rocfft_execute`.  This function takes
+#. Execute the plan by calling :cpp:func:`rocfft_execute`.  This function accepts
```
randyh62: This function accepts...

The change from _takes_ to _accepts_ seems to imply that it will accept but not require arrays of pointers. Does it take something other than arrays of pointers? I am just concerned that simply by changing takes to accept you have not actually communicated anything.

docs/how-to/working-with-rocfft.rst
```diff
@@ -353,8 +356,8 @@ The workflow for fields is as follows:
    added to the field (via calls to :cpp:func:`rocfft_field_add_brick`), and
    must point to memory on the device that was specified at that time.
 
-   For in-place transforms, only pass input pointers and do not pass output
-   pointers.
+   For in-place transforms, only pass input pointers and do not pass
```
randyh62: I would turn this into an important note:

.. important::
   For in-place transforms, only pass input pointers and do not pass...

docs/how-to/working-with-rocfft.rst
randyh62: I see from the next couple of paragraphs, that the accepting of arrays of pointers for input and output specifically applies to  distributed transforms. I would make that more clear: 

This function accepts arrays of pointers. For in-place transforms, the arrays of pointers should be provided for the inputs only. For distributed transforms, the arrays of pointers include both inputs and outputs.

docs/what-is-rocfft.rst
```diff
@@ -22,6 +22,11 @@ The rocFFT library provides a fast and accurate platform for calculating discret
 * Real and complex FFTs
 * Arbitrary lengths, with optimizations for combinations of powers of 2, 3, 5, 7, 11, 13, and 17
 
+rocFFT also has experimental support for:
+
+* Distributing transforms across multiple GPU devices in a single process
```
randyh62, line 27: Here we are introducing the idea of distributed transforms, which are new and experimental, but we have not actually discussed in-place transforms, which have existed but are not discussed here.  Maybe we should briefly discuss the difference between in-place and distributed transforms prior to this?

randyh62 (approved):
I have provided some comments, but otherwise it seems good to me.

docs/what-is-rocfft.rst
evetsso: Ah, I think you're thinking that "in-place" is the opposite of "distributed", but that's not the case.

"In-place" means the output of an operation is written to the same memory its input was read from.  The opposite of that is "out-of-place", which means the output is written to different memory as the input.

"Distributed" means multiple GPU devices participate in the operation.  We'd call the opposite of that "single-device".

It's reasonable to have any combination of these options, i.e.:
* in-place, single-device FFT
* out-of-place, single-device FFT
* in-place, distributed FFT
* out-of-place, distributed FFT

We can look at documenting this better, but it sounds separate from what we're adding here.

docs/how-to/working-with-rocfft.rst
evetsso: I've noticed a mis-wording in this paragraph and rewritten it.  There's a couple of other places where we could restructure this kind of text into notes so I can look at that in develop.

docs/how-to/working-with-rocfft.rst
evetsso: I've reverted the takes -> accepts change.  I think I started out rewriting that sentence but then changed my mind and didn't quite undo everything.

The rest I think is only a concern when "distributed" is treated as the opposite of "in-place".

PR #530, opened 2024-10-20
Require rocFFT use hip-config.cmake to find HIP

littlewu2508:
Sometimes rocFFT use FindHIP.cmake which is not desired -- it does not have hip::device and hip::host target imported and result in cmake generating error.

Reference: https://bugs.gentoo.org/932155


evetsso (approved):

cgmb:
> \[FindHIP.cmake\] does not have hip::device and hip::host target imported and result in cmake generating error.

There were some discussions ages ago about making FindHIP.cmake do a `find_package(hip CONFIG)` itself to fix this, but I guess the discussion never ended up leading to action. We should probably enhance FindHIP to eliminate this problem for all libraries and programs that use HIP.

malcolmroberts (approved):

eng-flavio-teixeira (approved):

PR #535, opened 2024-11-26
Add graph-safe reference

neon60:
Add reference to the graph safe section.


amd-jnovotny (approved):
looks good from the docs perspective

PR #544, opened 2025-02-11
Normalizes dependencies and fix build RPATH.

stellaraccident:
* Switches from -ldl and pthread to more modern CMake constructs (removing WIN32 carveout).
* Uses find_package to locate hiprtc and depends on it as an imported library.
* Removes BUILD_WITH_INSTALL_RPATH as this results in a broken build layout (this option should only be used in rare cases where the install RPATH is origin-relative *and* the build and install directory layout is the same, which it is not here).
* The above allowed the LD_LIBRARY_PATH hack to be removed because CMake build rpaths are always set to an absolute path by default, making them usable from the build tree without further intervention.


evetsso:
PRs need to target the default branch (`develop`), not `mainline`.  The hipRTC change is already present in `develop` as well.

PR #556, opened 2025-04-23
Add partial-pass code generator for 3D FFTs

eng-flavio-teixeira:
First piece for getting partial-pass working with arbitrary 3D FFTs.

This PR replaces the hardcoded 64x64x64 partial-pass SBRR and SBCC kernels (rtc_partial_pass_sbrr_64_64_64.cpp and rtc_partial_pass_sbcc_64_64_64.cpp) with kernel code generators.

There are still hardcoded values for 64x64x64 in a few other places, and these will be addressed once the kernels are fully configurable in kernel-generator.py. 

Some of the offset calculations in the new code generators (calculate_offsets() methods) may need to be changed once I can easily test and validate different lengths with the upcoming PR that address the configurability of 3D partial-pass kernels.


library/src/rtc_stockham_gen.cpp
```diff
@@ -297,10 +314,21 @@ std::string stockham_rtc(const StockhamGeneratorSpecs& specs,
     {
         std::unique_ptr<StockhamKernel> kernel;
         if(scheme == CS_KERNEL_STOCKHAM)
-            kernel = std::make_unique<StockhamKernelRR>(specs);
+        {
+            if(ppType == PartialPassType::PPT_SBRR)
```
malcolmroberts: Aren't there three cases here?  A switch statement would defend against missed options.

library/src/rtc_stockham_gen.cpp
```diff
@@ -315,9 +343,50 @@ std::string stockham_rtc(const StockhamGeneratorSpecs& specs,
...
+                = std::make_unique<Function>(kernel_pp->generate_lds_from_reg_output_pp_function());
+            lds2reg_pp_steps = std::make_unique<Function>(
+                kernel_pp->generate_lds_to_reg_input_step_3_4_function());
+            reg2lds_pp_steps = std::make_unique<Function>(
+                kernel_pp->generate_lds_from_reg_output_pp_step_3_4_function());
+            local_transpose_pp
+                = std::make_unique<Function>(kernel_pp->generate_local_transpose_pp_function());
+            device = std::make_unique<Function>(kernel_pp->generate_device_function());
+            break;
+        }
```
malcolmroberts, line 364: default case should raise an exception

library/src/device/generator/stockham_pp_gen_cc.h
```diff
@@ -0,0 +1,998 @@
...
+
+// TODO: Once partial pass is fully configurable in kernel-generator.py:
+//      - Test with factors_pp.size() > 1.
+//      - Revisit all usages of transforms_per_block_pp and threads_per_transform.
+//      - Different input/output strides.
+//      - Revisit mod 128 usage in calculate_offsets() with different input lengths,
+//        (logic is required to work with nbatch > 1)
+//      - Revisit factor 192 logic in calculate_offsets() with different input lengths
+//      - Revisit and test local transpose logic for different input lengths
+
```
malcolmroberts, line 32: I would really appreciate some documentation here; it's hard to tell why these functions are needed, and how they fit into the larger picture,

library/src/rtc_stockham_gen.cpp
eng-flavio-teixeira: Thanks, I've reorganized the code and now the three cases are addressed.

library/src/rtc_stockham_gen.cpp
eng-flavio-teixeira: Thanks, new commit should address this.

library/src/device/generator/stockham_pp_gen_cc.h
eng-flavio-teixeira: I've added some comments for both generators, and that should hopefully clarify their purposes.

malcolmroberts (approved):

evetsso (approved):

PR #557, opened 2025-04-24
add single-precision kernels for lengths 4704, 5488, 6144, 6561, 8192

evetsso:
64kiB LDS is still plenty to implement some larger 1D lengths, for single-precision.

Tested on gfx900, 908, 90a, 942, 1030, 1100, 1102 and it looks good all around.  The heuristic looks conservative enough to be safe.

Speedups range from 1.11 for a length like 14348907 (which uses 6561 as a sub-problem) to over 2x on large batch lengths like 4704 and 6144, on some architectures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ Documentation for rocFFT is available at
 
 ## rocFFT 1.0.34 (unreleased)
 
+### Optimized
+
+* Implemented single-precision 1D kernels for lengths:
+  - 4704
+  - 5488
+  - 6144
+  - 6561
+  - 8192
+
 ### Removed
 
 * Removed rocfft-rider legacy compatibility from clients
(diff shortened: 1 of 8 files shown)


amd-jnovotny (approved):
Changelog OK

eng-flavio-teixeira (approved):
LG. Are there any updates to the performance numbers, or the ones you've mentioned previously (speed up 1.1x to over 2x) still hold true?

evetsso:
> LG. Are there any updates to the performance numbers, or the ones you've mentioned previously (speed up 1.1x to over 2x) still hold true?

Still true.  The last commit I made does improve things a little bit more for a couple of lengths (wgs > 256 starts to pay off for longer 1D lengths) but no fundamental changes in the numbers.

af-ayala (approved):

Kevonosdiaz (approved):

malcolmroberts (approved):

PR #562, opened 2025-04-29
Perform global transpose using all-to-all whenever possible

af-ayala:
Currently, the global transpose operation uses `Alltoallv` communication by default. However, `Alltoall` is generally faster and more efficient across various MPI distributions. In this PR, we propose using all-to-all communication wherever possible, and falling back to `Alltoallv` when necessary.

### DIFF
--- a/library/src/include/tree_node.h
+++ b/library/src/include/tree_node.h
@@ -1288,21 +1288,37 @@ private:
 // The former is preferable, as it is usually more optimized.
 struct CommAllToAll : public MultiPlanItem
 {
-    CommAllToAll() = default;
+    CommAllToAll(rocfft_precision           _precision,
+                 rocfft_array_type          _arrayType,
+                 const std::vector<size_t>& _sendOffsets,
+                 const std::vector<size_t>& _sendCounts,
+                 const std::vector<size_t>& _recvOffsets,
+                 const std::vector<size_t>& _recvCounts,
+                 BufferPtr                  _sendBuf,
+                 BufferPtr                  _recvBuf)
+        : precision(_precision)
+        , arrayType(_arrayType)
+        , sendOffsets(_sendOffsets)
+        , sendCounts(_sendCounts)
+        , recvOffsets(_recvOffsets)
+        , recvCounts(_recvCounts)
+        , sendBuf(_sendBuf)
+        , recvBuf(_recvBuf)
+    {
+        // Currently MPI interface uses 32-bit signed ints, so assert
+        // that our counts/offsets don't overflow that type
+        auto checkArray = [](const std::vector<size_t>& arr) {
+            if(std::any_of(arr.begin(), arr.end(), [](size_t elem) {
+                   return elem > std::numeric_limits<int>::max();
+               }))
+                throw std::runtime_error("count/offset exceeds int max");
+        };
 
-    rocfft_precision  precision;
-    rocfft_array_type arrayType;
-
-    // counts and offsets are all in elements (where element size is
-    // knowable from precision + array type), for the current rank
-    std::vector<size_t> sendOffsets;
-    std::vector<size_t> sendCounts;
-    std::vector<size_t> recvOffsets;
-    std::vector<size_t> recvCounts;
-
-    // send/receive buffers
-    BufferPtr sendBuf;
-    BufferPtr recvBuf;
+        checkArray(sendOffsets);
+        checkArray(sendCounts);
+        checkArray(recvOffsets);
+        checkArray(recvCounts);
+    }
 
     // enum for error handling for different multi-process communication
     // libraries such as MPI, RCCL, rocSHMEM, etc.
@@ -1335,6 +1351,21 @@ struct CommAllToAll : public MultiPlanItem
         // runs on all ranks
         return true;
     }
+
+private:
+    rocfft_precision  precision;
+    rocfft_array_type arrayType;
+
+    // counts and offsets are all in elements (where element size is
+    // knowable from precision + array type), for the current rank
+    std::vector<size_t> sendOffsets;
+    std::vector<size_t> sendCounts;
+    std::vector<size_t> recvOffsets;
+    std::vector<size_t> recvCounts;
+
+    // send/receive buffers
+    BufferPtr sendBuf;
+    BufferPtr recvBuf;
 };
 
 // Tree-structured FFT plan.  This is specific to a single device on
--- a/library/src/include/tree_node.h
+++ b/library/src/include/tree_node.h
@@ -1284,9 +1284,11 @@ private:
 // Send data from all ranks to all ranks in the plan.  Each rank must
 // send from/to a single buffer (with different read/write offsets
 // for each other rank).
-struct CommAllToAllv : public MultiPlanItem
+// The all-to-all communication is performed using MPI_Ialltoall or MPI_Ialltoallv.
+// The former is preferable, as it is usually more optimized.
+struct CommAllToAll : public MultiPlanItem
 {
-    CommAllToAllv() = default;
+    CommAllToAll() = default;
 
     rocfft_precision  precision;
     rocfft_array_type arrayType;
@@ -1307,6 +1309,7 @@ struct CommAllToAllv : public MultiPlanItem
                       void*                 out_buffer[],
                       rocfft_execution_info info,
                       size_t                multiPlanIdx) override;
+
     void Wait() override;
 
     void Print(rocfft_ostream& os, const int indent) const override;
--- a/library/src/include/tree_node.h
+++ b/library/src/include/tree_node.h
@@ -1061,19 +1061,19 @@ struct CommPointToPoint : public MultiPlanItem
     }
 
 private:
-    rocfft_precision  precision;
-    rocfft_array_type arrayType;
+    const rocfft_precision  precision;
+    const rocfft_array_type arrayType;
 
     // number of elements to copy
-    size_t numElems;
+    const size_t numElems;
 
-    rocfft_location_t srcLocation;
-    BufferPtr         srcPtr;
-    size_t            srcOffset = 0;
+    const rocfft_location_t srcLocation;
+    const BufferPtr         srcPtr;
+    const size_t            srcOffset = 0;
 
-    rocfft_location_t destLocation;
-    BufferPtr         destPtr;
-    size_t            destOffset = 0;
+    const rocfft_location_t destLocation;
+    const BufferPtr         destPtr;
+    const size_t            destOffset = 0;
     // Stream to run the async operation in
     hipStream_wrapper_t stream;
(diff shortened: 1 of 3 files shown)


malcolmroberts:
```[ 34%] Built target rocfft-function-pool
[ 37%] Built target rocfft_aot_helper
[ 39%] Built target rocfft_kernel_config_search
[ 40%] Building CXX object library/src/CMakeFiles/rocfft.dir/plan.cpp.o
/users/mroberts/repo/rocfft1/library/src/plan.cpp:2270:60: error: unknown type name 'CommAllToAllv'; did you mean 'CommAllToAll'?
    auto alltoall_ptr                   = std::make_unique<CommAllToAllv>();
                                                           ^~~~~~~~~~~~~
                                                           CommAllToAll
/users/mroberts/repo/rocfft1/library/src/include/tree_node.h:1289:8: note: 'CommAllToAll' declared here
struct CommAllToAll : public MultiPlanItem
       ^
1 error generated when compiling for gfx90a.
make[2]: *** [library/src/CMakeFiles/rocfft.dir/build.make:90: library/src/CMakeFiles/rocfft.dir/plan.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:691: library/src/CMakeFiles/rocfft.dir/all] Error 2
make: *** [Makefile:156: all] Error 2
```
This doesn't compile on Lumi.

malcolmroberts (changes requested):
Needs work compiling on Lumi

af-ayala:
@malcolmroberts 
It seems you tried a previous commit, 
Please use the latest one [ed8f32f](https://github.com/ROCm/rocFFT/pull/562/commits/ed8f32fdf4d41a6280d42cd586a7bc1daaac6061), where no reference to CommAlltoAllv is done in the plan.cpp file.

library/src/tree_node.cpp
```diff
@@ -894,72 +894,119 @@ void CommGather::Print(rocfft_ostream& os, const int indent) const
...
 #ifdef ROCFFT_MPI_ENABLE
 
-    // MPI takes ints for everything, convert our size_t elements to int bytes
-    auto convertToInt = [](const std::vector<size_t>& src, std::vector<int>& dest) {
-        dest.reserve(src.size());
-        for(auto i : src)
+    const auto elem_size = element_size(precision, arrayType);
+
+    // check if uniform counts
+    bool   uniform_counts = true;
```
evetsso: You could use the standard library to work this out, with something like:
```c++
auto count_matches_first = [&](size_t count) { return count == sendCounts[0];};
bool uniform_counts = std::all_of(sendCounts.begin(), sendCounts.end(), count_matches_first) || std::all_of(recvCounts.begin(), recvCounts.end(), count_matches_first);
```

evetsso (approved):
Can you provide any benchmark numbers to show what the improvement is?

malcolmroberts (changes requested):
please provide a few benchmarks; it would be good to see what kind of performance improvement we get.

af-ayala:
We can observe over ~20% improvement for sizes typically used by applications relying on 3D FFTs.

Detailed data here:
https://amdcloud-my.sharepoint.com/:w:/g/personal/alaayala_amd_com/EaQ9KTrM2-hOl_7_Zuaz2gEBJvsemNHFlx50Nnkaxd-VLg?e=XfnTTO

eng-flavio-teixeira (approved):

library/src/tree_node.cpp
```diff
@@ -894,72 +894,112 @@ void CommGather::Print(rocfft_ostream& os, const int indent) const
...
-    if(mpiret != MPI_SUCCESS)
-        throw std::runtime_error("MPI_Ialltoallv failed: " + std::to_string(mpiret));
+
+    if(uniform_counts)
+    {
+        if(LOG_PLAN_ENABLED())
+            log_plan("Using MPI_Ialltoall\n");
+
+        if(sendCounts[0] * elem_size > static_cast<size_t>(std::numeric_limits<int>::max()))
+            throw std::runtime_error(
```
malcolmroberts: Don't throw an exception before a collective MPI call.

library/src/tree_node.cpp
```diff
@@ -894,72 +894,112 @@ void CommGather::Print(rocfft_ostream& os, const int indent) const
...
+        if(LOG_PLAN_ENABLED())
+            log_plan("Using MPI_Ialltoallv\n");
+
+        // MPI takes ints for everything, convert our size_t elements to int bytes
+        auto convertToInt = [](const std::vector<size_t>& src, std::vector<int>& dest) {
+            dest.reserve(src.size());
+            for(auto i : src)
+            {
+                if(i > std::numeric_limits<int>::max())
+                    throw std::runtime_error("MPI integer limit exceeded");
```
malcolmroberts: Can't throw here either; if one rank throws, then we deadlock.

malcolmroberts (dismissed):
We need to handle error cases without ending up in an MPI deadlock.

library/src/tree_node.cpp
```diff
@@ -894,72 +894,112 @@ void CommGather::Print(rocfft_ostream& os, const int indent) const
...
-    std::vector<int> intRecvOffsets;
-    std::vector<int> intRecvCounts;
-    convertToInt(sendOffsets, intSendOffsets);
-    convertToInt(sendCounts, intSendCounts);
-    convertToInt(recvOffsets, intRecvOffsets);
-    convertToInt(recvCounts, intRecvCounts);
+    // check if uniform counts
+    auto count_matches_first = [&](size_t count) { return count == sendCounts[0]; };
+    bool uniform_counts = std::all_of(sendCounts.begin(), sendCounts.end(), count_matches_first)
+                          || std::all_of(recvCounts.begin(), recvCounts.end(), count_matches_first);
```
Kevonosdiaz: I was under the impression that all of the send and receive counts should be the same, so I thought this should be an `&&` of the two instead of an `||` operation. If not, could you explain this?

evetsso: Good catch.  In my defense, I said "something like" in my suggestion.  :)

But this really means we need to add testcases for the Alltoallv as well.

library/src/tree_node.cpp
```diff
@@ -894,72 +894,127 @@ void CommGather::Print(rocfft_ostream& os, const int indent) const
...
+                                          send_count_bytes,
+                                          MPI_CHAR,
+                                          plan->desc.mpi_comm,
+                                          &request);
+
+        if(mpiret != MPI_SUCCESS)
+        {
+            std::cerr << "MPI_Ialltoall failed on rank " << local_comm_rank
+                      << " with error: " << mpiret << std::endl;
+            MPI_Abort(plan->desc.mpi_comm, mpiret);
```
malcolmroberts: MPI_Abort isn't a solution either.  The collective call has failed, which should mean that it failed on all ranks in the communicator, and now we should return a status code which indicates that things didn't work in the library.  MPI_Abort will exit the process, which stops us from being able to log debug information.

What we should do is get to the collective MPI call, make sure that it fails, then return an error status code.

library/src/tree_node.cpp
```diff
@@ -973,12 +1053,28 @@ void CommAllToAllv::Print(rocfft_ostream& os, const int indent) const
...
-    os << indentStr << "CommAllToAllv " << precision_name(precision) << " "
-       << PrintArrayType(arrayType) << ":\n";
-    printVec("sendOffsets", sendOffsets);
-    printVec("sendCounts", sendCounts);
-    printVec("recvOffsets", recvOffsets);
-    printVec("recvCounts", recvCounts);
+    // determine whether counts are uniform
+    auto count_matches_first = [&](size_t count) { return count == sendCounts[0]; };
+    bool uniform_counts = std::all_of(sendCounts.begin(), sendCounts.end(), count_matches_first)
+                          || std::all_of(recvCounts.begin(), recvCounts.end(), count_matches_first);
```
Kevonosdiaz: One more place to update to && instead of || for uniform_counts

af-ayala: It was the same as above, for some reason my tree messed up.
Thank you for catching this again.

Kevonosdiaz (approved):

library/src/tree_node.cpp
```diff
@@ -894,72 +894,152 @@ void CommGather::Print(rocfft_ostream& os, const int indent) const
...
+        // guard against overflow of MPI_INT
+        if(sendCounts[0] * elem_size > static_cast<size_t>(std::numeric_limits<int>::max()))
         {
-            if(i > std::numeric_limits<int>::max())
-                throw std::runtime_error("MPI integer limit exceeded");
-            dest.push_back(i);
+            comm_status   = COMM_MPI_ERROR;
+            error_message = "Rank " + std::to_string(local_comm_rank)
+                            + ": element size * count_per_rank exceeds MPI_INT limit";
+            return;
```
malcolmroberts: If we return on just one rank, isn't this going to produce a hang?

af-ayala: We can move it to the plan creation. I am also sharing some notes about error handling for collective MPI.

library/src/tree_node.cpp
```diff
@@ -973,12 +1053,28 @@ void CommAllToAllv::Print(rocfft_ostream& os, const int indent) const
...
-    printVec("sendCounts", sendCounts);
-    printVec("recvOffsets", recvOffsets);
-    printVec("recvCounts", recvCounts);
+    // determine whether counts are uniform
+    auto count_matches_first = [&](size_t count) { return count == sendCounts[0]; };
+    bool uniform_counts = std::all_of(sendCounts.begin(), sendCounts.end(), count_matches_first)
+                          && std::all_of(recvCounts.begin(), recvCounts.end(), count_matches_first);
+
+    os << indentStr << "CommAllToAll " << precision_name(precision) << " "
+       << PrintArrayType(arrayType) << (uniform_counts ? " (MPI_Ialltoall)" : " (MPI_alltoallv)")
```
regan-amd:
```suggestion
       << PrintArrayType(arrayType) << (uniform_counts ? " (MPI_Ialltoall)" : " (MPI_Ialltoallv)")
```
isn't it?

library/src/include/tree_node.h
```diff
@@ -1335,6 +1351,21 @@ struct CommAllToAll : public MultiPlanItem
...
+private:
+    rocfft_precision  precision;
+    rocfft_array_type arrayType;
+
+    // counts and offsets are all in elements (where element size is
+    // knowable from precision + array type), for the current rank
+    std::vector<size_t> sendOffsets;
+    std::vector<size_t> sendCounts;
+    std::vector<size_t> recvOffsets;
+    std::vector<size_t> recvCounts;
```
regan-amd: Can these members be modified after initialization or could we `const`-qualify all of them (given how they're initialized)?

PR #563, opened 2025-04-29
Handle case variances for command-line boolean options in rocfftslurmtest

malcolmroberts:


Kevonosdiaz:
Alternatively could use `action='store_true'` to make it a flag instead of requiring an argument after. Also looks like yapf in static analysis is complaining :)

malcolmroberts:
> Alternatively could use `action='store_true'` to make it a flag instead of requiring an argument after. Also looks like yapf in static analysis is complaining :)

I would prefer to do that, but then it doesn't work with the config file, which expects an actual value after the command line option.  Python's argparse and configparser libraries don't really talk well with eachother.

Kevonosdiaz (approved):

eng-flavio-teixeira (approved):

PR #565, opened 2025-05-01
generator: store LDS size with each kernel config

evetsso:
This makes it clearer which size a config is for, rather than trying to work that out indirectly by trying to infer how much LDS a config would use.

Originally I'd hoped that 4dc5d94daec4d4b20789eed3e101472ff8ea2e68 would be enough to handle additional LDS sizes but it's a lot more straightforward if every kernel config just explicitly knows that size it's for.

### DIFF
--- a/library/src/device/generator.py
+++ b/library/src/device/generator.py
@@ -811,13 +811,10 @@ class Map(BaseNodeOps):
             Throw('std::runtime_error("' + str(what_error) + '")'))
         return If(Equal(status, "false"), throw)
 
-    def assert_insert(self, key, value, def_key_pool, function_map):
-        insert = Call('insert_default_entry',
-                      arguments=ArgumentList(key, value, def_key_pool,
-                                             function_map)).inline()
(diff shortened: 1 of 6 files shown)


malcolmroberts (approved):

eng-flavio-teixeira (approved):

af-ayala (approved):

malcolmroberts:
So, this is perhaps a tricky question, but if we re-tune, does this give any performance regressions?

evetsso:
> So, this is perhaps a tricky question, but if we re-tune, does this give any performance regressions?

I'm not sure what you mean - this isn't changing any kernel configs.  It's just making the existing kernel configs explicitly say they are for 64kiB LDS.  Architectures that have more would prefer configs for their specific LDS size, but will use configs for smaller LDS if that's all that's available.

Adding configs for large LDS when we already have 64kiB configs could cause performance changes, but that's not what this PR is doing.

PR #570, opened 2025-05-05
Create test tokens starting at local gpu index 0

malcolmroberts:


evetsso (approved):

shared/fft_params.h
```diff
@@ -2285,12 +2285,12 @@ class fft_params
     // starts with batch dimension, followed by FFT dimensions
     // slowest to fastest.
     // num_nodes represents the number of nodes used in the parallel
-    // computer, which are assumed to have at least ngpus each
-    void distribute_field(int                              ngpus,
+    // computer, which are assumed to have at least gpusperrank each
+    void distribute_field(int                              gpusperrank,
```
regan-amd, line 2289: Purely aesthetic and I'm not aware (yet) of any syntax/naming conventions: wouldn't `gpus_per_rank` be better aligned with how other arguments are named?

shared/fft_params.h
```diff
@@ -2285,12 +2285,12 @@ class fft_params
     // starts with batch dimension, followed by FFT dimensions
     // slowest to fastest.
     // num_nodes represents the number of nodes used in the parallel
```
regan-amd:
```suggestion
    // mpi_size represents the number of MPI ranks being used
```

evetsso: Currently, our multi-process implementation uses MPI.  We hope to add support for other backends that are not MPI, so that's the reason the wording is more generic.

That said, it's entirely possible that those other backends will still use the word "rank" to describe each process anyway.

regan-amd: Thanks! The motivation for my suggestion was to update the part of the explanatory comment that still refers to `num_nodes` albeit that variable is renamed below in this pull request (to maintain consistency between the declaration of `distribute_field` and its explanatory comment).
Though "MPI ranks" might not be future-proof in light of what you point out, it also seemed to me that the renamed variable (`num_nodes` -> `mpi_size`) no longer captures the "number of nodes" in this context, given other changes (_e.g._, removal of `get_num_nodes`).

regan-amd (approved):

PR #571, opened 2025-05-06
generator: add 1D configs for 160kiB LDS

evetsso:
New lengths (single- and double- precision):
- 4704
- 5488
- 6144
- 6561
- 8192
- 9216
- 10000
- 10240

Single-precision only:
- 10752
- 11200
- 12288
- 16384
- 16807
- 18816
- 19200
- 20480

[figs-gfx950.zip](https://github.com/user-attachments/files/20065884/figs-gfx950.zip)

Attached some performance figures, looks like this is a pretty easy win.

### DIFF
--- a/library/src/device/kernel-generator.py
+++ b/library/src/device/kernel-generator.py
@@ -241,7 +241,8 @@ def kernel_name(ns):
     return f'rocfft_len{length}{postfix}'
 
 
-LDS_160k = 160*1024
+LDS_160k = 160 * 1024
+
 
 # yapf: disable
 def list_small_kernels():
--- a/library/src/device/kernel-generator.py
+++ b/library/src/device/kernel-generator.py
@@ -241,6 +241,8 @@ def kernel_name(ns):
     return f'rocfft_len{length}{postfix}'
 
 
+LDS_160k = 160*1024
+
 # yapf: disable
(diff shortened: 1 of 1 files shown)


eng-flavio-teixeira:
Do you have any plans for dealing with:

**NodeFactory::Map1DLength const NodeFactory::map1DLengthSingle 
NodeFactory::Map1DLength const NodeFactory::map1DLengthDouble**

for larger LDS sizes? 

Those data structures seem to be tailored to 64KiB LDS, and do not take into account the newer sizes. They would most likely have different entries for 160KiB.

evetsso:
> Those data structures seem to be tailored to 64KiB LDS, and do not take into account the newer sizes. They would most likely have different entries for 160KiB.

I don't think we need to do anything for now.  Yes,  it's tailored for 64kiB LDS, but if we have a kernel for a 1D length, then we use that and don't use these maps.  If there's a length that's just too big for larger LDS, then IMO the decomposition in the map is still best - trying to use more LDS for one of the kernels would make the decomposition lopsided.

malcolmroberts (approved):

eng-flavio-teixeira (approved):

Kevonosdiaz (approved):

PR #572, opened 2025-05-08
Add support for fully configurable partial-pass 3D kernels

eng-flavio-teixeira:
Add the ability to fully configure partial-pass 3D kernels in kernel-generator.py. This should remove the need for the remaining hard-coded 64x64x64 partial-pass parameters across the library.

The main changes include:
 - kernel-generator.py now has a list of CS_3D_PP kernels, where partial-pass parameters are fully configurable. 
 - Stockham_gen and kernel-generator.py interaction mechanism via JSON needed to understand the concept of linked partial-pass kernels.
 - Function pool and function map key needed a new concept for the linked partial-pass kernels, which are now queryable in the pool.
 - New compute schemes specifically added for partial-pass (CS_3D_PP, CS_STOCKHAM_PP, and CS_STOCKHAM_PP_BLOCK_CC). These should simplify handling of nodes in the plan tree structure, and make them clearly distinct from CS_3D_RC nodes.


library/src/include/function_map_key.h
```diff
@@ -459,4 +476,82 @@ struct SimpleHash
     }
 };
```
malcolmroberts, line 482: It's not clear from context what these new structs are for; would it be possible to add some code comments, please?

eng-flavio-teixeira: Sure, new commit is in with the comments.

Kevonosdiaz:
Haven't looked much into this PR yet, but it might be a good time to update the default value of the variable at 
https://github.com/ROCm/rocFFT/blob/dcfff3f35765f4d8c5fad66cbeac3fb4bd7bd861/CMakeLists.txt#L121 
to 16 or higher if the function pool is growing, since it might improve build times a bit.

malcolmroberts (approved):

eng-flavio-teixeira:
> Haven't looked much into this PR yet, but it might be a good time to update the default value of the variable at
> 
> https://github.com/ROCm/rocFFT/blob/dcfff3f35765f4d8c5fad66cbeac3fb4bd7bd861/CMakeLists.txt#L121
> 
> 
> to 16 or higher if the function pool is growing, since it might improve build times a bit.

Did you see any noticeable improvement going from 8 to 16? This PR does not add more kernels to the function pool, but the next one will (but not an unreasonable amount).

Kevonosdiaz:
> Did you see any noticeable improvement going from 8 to 16? This PR does not add more kernels to the function pool, but the next one will (but not an unreasonable amount).

It slowed down my build time by a couple seconds when going from 8 to 16, and building just the `rocfft-test` target with ninja, and without ccache. When there are more kernels added then maybe it'll be worth checking again.

evetsso (approved):

eng-flavio-teixeira:
> > Did you see any noticeable improvement going from 8 to 16? This PR does not add more kernels to the function pool, but the next one will (but not an unreasonable amount).
> 
> It slowed down my build time by a couple seconds when going from 8 to 16, and building just the `rocfft-test` target with ninja, and without ccache. When there are more kernels added then maybe it'll be worth checking again.

Sounds good, I will keep that in mind for the next PR.

Kevonosdiaz (approved):

af-ayala (approved):

PR #575, opened 2025-05-14
Improve kernel io log precision

eng-flavio-teixeira:
External user scripts that rely on kernel io logs with ROCFFT_LAYER=16 can have trouble with the precision of the logged data. 

For example, if an external script tries to compare the kernel ouput data in double-precision with another FFT implementation, accuracy comparisons may fail because the log is not being printed out with the required precision.

This PR changes printbuffer to use the actual precision that the input/output data was generated/computed, and change the data stream to print with a predefined number of decimals.

### DIFF
--- a/library/src/powX.cpp
+++ b/library/src/powX.cpp
@@ -335,19 +335,19 @@ void DebugPrintBuffer(rocfft_ostream&             stream,
         case rocfft_precision_half:
         {
             buffer_printer<rocfft_fp16> s;
-            s.print_buffer(bufvec, length_rm, stride_rm, batch, dist, print_offset, stream);
+            s.print_buffer_half(bufvec, length_rm, stride_rm, batch, dist, print_offset, stream);
             break;
         }
         case rocfft_precision_single:
         {
             buffer_printer<float> s;
-            s.print_buffer(bufvec, length_rm, stride_rm, batch, dist, print_offset, stream);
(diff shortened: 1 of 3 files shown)


Kevonosdiaz (approved):

evetsso:
This seems harmless enough, but I do wonder if LAYER=16 is the right tool for the job if you're finding you need high precision when comparing results.

A while back I mused about some way to just directly write out the buffers to disk rather than try to make the human-readable.  Then you could have some code that directly worked with the floating-point values rather than try to translate them to/from strings.

eng-flavio-teixeira:
> This seems harmless enough, but I do wonder if LAYER=16 is the right tool for the job if you're finding you need high precision when comparing results.
> 
> A while back I mused about some way to just directly write out the buffers to disk rather than try to make the human-readable. Then you could have some code that directly worked with the floating-point values rather than try to translate them to/from strings.

For context, the user script in question is in Octave/MATLAB, and I've added it to the rocFFT-misc repo if you're interested in taking a look. 

It may be possible to import binary data directly into Octave if we were to write out the buffers from the library, but I need to confirm if this would work with all the scripts already implemented without major changes.

I agree that parsing the log from the script is not very nice, but that is already implemented and working. However, if something changes in the log layout, then it could potentially break the parsing script, so not the most robust implementation.

evetsso (approved):

PR #578, opened 2025-05-20
node_factory: only use pow2 kernels to decompose pow2 lengths

evetsso:
Fixes a problem where pow2 decomposition code was incorrectly assuming the largest 1D length in the pool is also pow2.

### DIFF
--- a/library/src/node_factory.cpp
+++ b/library/src/node_factory.cpp
@@ -658,7 +658,11 @@ ComputeScheme NodeFactory::Decide1DScheme(const function_pool& pool, NodeMetaDat
         }
         else
         {
-            auto largest = pool.get_largest_length(nodeData.precision);
+            // get largest pow2 1D length
+            auto pow2_lengths = pool.get_lengths(
+                nodeData.precision, CS_KERNEL_STOCKHAM, [](size_t len) { return IsPo2(len); });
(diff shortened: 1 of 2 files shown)


eng-flavio-teixeira (approved):

library/src/node_factory.cpp
```diff
@@ -690,7 +690,11 @@ ComputeScheme
         }
         else
         {
-            auto largest = pool.get_largest_length(nodeData.precision);
+            // get largest pow2 1D length
+            auto pow2_lengths = pool.get_lengths(
+                nodeData.precision, CS_KERNEL_STOCKHAM, [](size_t len) { return IsPo2(len); });
+            auto largest = *std::max_element(pow2_lengths.cbegin(), pow2_lengths.cend());
```
regan-amd: Is there any risk of `pow2_lengths` being empty? Should that be checked against before dereferencing the result of `std::max_element`?

evetsso: It's hard to think of why we'd be in a situation where we don't even have a single pow2 length available, but I've added the check just in case.

regan-amd (approved):

malcolmroberts (approved):

Kevonosdiaz (approved):

PR #580, opened 2025-05-21
Move gpuidvar dereference to actual bash in slurm test infrastructure

malcolmroberts:


evetsso (approved):
It looks to me like dereferencing the variable is being moved to `bash` instead of python?

eng-flavio-teixeira (approved):

PR #581, opened 2025-05-21
bluestein: only look for pow2 lengths for single-kernel

evetsso:
Randomized tests noticed that real_forward_len_9409_single_ip_batch_1_istride_1_R_ostride_1_HI_idist_9410_odist_4705_ioffset_0_0_ooffset_0_0 broke, because single-kernel Bluestein was also trying to use `function_pool::get_largest_length` and assuming the largest length was pow2.  But since 9e97f280762b12c6e7085f614ccf19a05aee9df3 and this are the only things that need `function_pool::get_largest_length`, I've instead renamed that to `get_largest_pow2_length` and used it in both places.

### DIFF
--- a/library/src/tree_node_bluestein.cpp
+++ b/library/src/tree_node_bluestein.cpp
@@ -95,6 +95,9 @@ BluesteinType BluesteinNode::DecideBlueType()
         return type;
     }
 
+    // Handle lengthBlue with its own non-fused FFT sub-plan.  This
+    // can be a multi-kernel L1D plan, or a 1-kernel Stockham plan if
+    // lengthBlue can be done with a single non-pow2 kernel.
     if(scheme == CS_L1D_CRT || scheme == CS_L1D_TRTRT || scheme == CS_KERNEL_STOCKHAM)
         return BluesteinType::BT_MULTI_KERNEL;
 
--- a/library/src/tree_node_bluestein.cpp
+++ b/library/src/tree_node_bluestein.cpp
@@ -95,7 +95,7 @@ BluesteinType BluesteinNode::DecideBlueType()
         return type;
     }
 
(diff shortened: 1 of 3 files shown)


library/src/tree_node_bluestein.cpp
```diff
@@ -95,7 +95,7 @@ BluesteinType BluesteinNode::DecideBlueType()
         return type;
     }
 
-    if(scheme == CS_L1D_CRT || scheme == CS_L1D_TRTRT)
+    if(scheme == CS_L1D_CRT || scheme == CS_L1D_TRTRT || scheme == CS_KERNEL_STOCKHAM)
```
eng-flavio-teixeira, line 101: Any specific reason for adding CS_KERNEL_STOCKHAM here?

Kevonosdiaz (approved):

malcolmroberts (approved):

af-ayala (approved):

eng-flavio-teixeira (approved):

library/src/tree_node_bluestein.cpp
evetsso: Yeah, this is necessary because we only use pow2 kernels for single-kernel Bluestein (due to their relatively low register use), but the largest 1D kernels we support are not necessarily pow2.

So it's possible now for `lengthBlue` to be small enough that we can do it with one kernel, and yet we still can't fuse all of Bluestein's algorithm into that kernel.  If that happens, then scheme here will be `CS_KERNEL_STOCKHAM`.

eng-flavio-teixeira: Makes sense, thanks. Perhaps adding a comment here would help as this is not that obvious at a first glance.

regan-amd (approved):

PR #584, opened 2025-05-26
generator: move EmbeddedType to specs

evetsso:
This allows the generator to know up front whether real-complex pre/post processing is needed, and avoid adding conditionals in the generated code.  Even though the conditions are constant, it's easier to not have to generate both arms of the conditional at the same time.

This *shouldn't* change anything performance-wise, since we're just moving from if/else clauses that check a static const variable to just the lines of code that should be running anyway if that variable has a given value.  But it's possible we're making the compiler's job slightly easier and that might improve things somehow.  I'll do full performance runs to check.

There's a bunch of other static const variables we currently define in a similar way that we could also get rid of.  I'm aiming for this one specifically because it'll help 2D_SINGLE kernels.


af-ayala (approved):

malcolmroberts:
I can reproduce the compilation errors that CI is hitting on this one.

evetsso:
> I can reproduce the compilation errors that CI is hitting on this one.

Right, CI (quite rightly) has `-DROCFFT_BUILD_OFFLINE_TUNER=ON` but I haven't added it to my default CMake command line.  I'll do that now.

Edit: misread the error, wasn't related to that option but the problem's fixed anyway.

eng-flavio-teixeira (approved):

malcolmroberts (approved):

malcolmroberts (approved):

library/src/rocfft_aot_helper.cpp
```diff
@@ -306,14 +306,6 @@ void build_stockham_function_pool(CompileQueue& queue)
                                                             {});
                 std::function<std::string(const std::string&)> generate_src
                     = [=](const std::string& kernel_name) -> std::string {
-                    StockhamGeneratorSpecs specs{factors,
```
Kevonosdiaz, line 309: I was just wondering why this section is being removed

evetsso: This variable is shadowing one that's earlier in the function.  This can cause us to compute the kernel's name with one specs, then potentially generate source code using a potentially different one.

Since kernel name is key to our kernel database, this causes bad things to happen.  So by removing this mostly-duplicated variable, we ensure the name + source are kept consistent the whole way through.

I'm not 100% sure why this didn't cause problems before this PR, but I didn't look into it that hard.

Kevonosdiaz: Yeah it seems pretty strange. I was just wondering if it was needed for some arcane reason maybe

Kevonosdiaz (approved):

PR #586, opened 2025-05-29
Add FFTW dependency for mpi workers when FFTW build was automatically triggered in cmake

malcolmroberts:


malcolmroberts:
This seems to resolve a build issue when FFTW isn't found and MPI is specified.

evetsso (approved):
Hmm..

malcolmroberts:
> Hmm..

Yeah, before this change, I had rocfft_mpi_worker being built before FFTW was finished building.  It then couldn't find  fftw3.h, and compilation failed.  This doesn't seem to happen all of the time - I suspect that it's more on systems where I'm building for the first time, which means that ccache can't really help.  Lots of build jobs seems to be a factor.  I'm also using make, so perhaps different behaviour with ninja.  (However, ninja is reportedly more liberal with build dependencies than make; either way, it has a different build DAG.)

evetsso:
Hah, I didn't actually mean to comment with `Hmm..` but the change does make sense.  There might be a better way to do it but this will work well enough.

af-ayala (approved):

PR #589, opened 2025-06-04
Make rocfft-test OpenMP linking public in cmake, since it is used at runtime

malcolmroberts:


malcolmroberts:
This resolves the issue where libomp isn't found when running rocfft-test (unless you update your LD_LIBRARY_PATH).

evetsso:
I'm not understanding how this helps - PUBLIC basically equals PRIVATE + INTERFACE.  But does adding INTERFACE scope change anything for executables?

For example, it doesn't look like this changes RUNPATH on the executable, so I don't see how this would help find libomp.so at runtime.

malcolmroberts:
> I'm not understanding how this helps - PUBLIC basically equals PRIVATE + INTERFACE. But does adding INTERFACE scope change anything for executables?
> 
> For example, it doesn't look like this changes RUNPATH on the executable, so I don't see how this would help find libomp.so at runtime.

Ok, it looks like I was looking at an older build; I can't reproduce the error that I was hitting.  Closing.

PR #590, opened 2025-06-04
Rename cmake variable ROCmCMakeBuildTools to ROCM

malcolmroberts:


malcolmroberts:
This handles a name change in the rocm build tools repo; seems to work better on some clusters now.

PR #595, opened 2025-06-10
sys_mem: count buffers as free, but not integrated device memory

evetsso:
Buffers are generally considered "available" memory as they can be easily dropped under memory pressure.  Also, both "total" and "free" memory on an integrated APU must account for "device" memory not being available to the host.

### DIFF
--- a/shared/sys_mem.h
+++ b/shared/sys_mem.h
@@ -115,7 +115,27 @@ private:
         if(sysinfo(&info) != 0)
             return;
         total_bytes = info.totalram * info.mem_unit;
-        free_bytes  = info.freeram * info.mem_unit;
+        free_bytes  = (info.freeram + info.bufferram) * info.mem_unit;
+
+        // Adjust memory numbers for APU
+        try
+        {
+            auto deviceProp = get_curr_device_prop();
+            // on integrated APU, we can't expect to reuse "device" memory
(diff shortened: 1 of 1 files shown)


malcolmroberts (approved):

shared/sys_mem.h
```diff
@@ -115,7 +115,27 @@ struct host_memory
...
+
+        // Adjust memory numbers for APU
+        try
+        {
+            auto deviceProp = get_curr_device_prop();
+            // on integrated APU, we can't expect to reuse "device" memory
+            // for "host" things.
+            if(deviceProp.integrated)
+            {
+                total_bytes -= deviceProp.totalGlobalMem;
```
regan-amd, line 128: Are cases where `deviceProp.totalGlobalMem > total_bytes` ruled out somehow (logically or otherwise)? If not, more care may be required for `total_bytes`, here above.

evetsso: Integrated GPUs take their `totalGlobalMem` from system memory, so it's difficult to imagine a situation where `total_bytes` would be any less than `deviceProp.totalGlobalMem`.  If that ever happened, I'd argue that's a bug in the runtime.

af-ayala (approved):

regan-amd (approved):

PR #598, opened 2025-06-13
sys_mem: check /proc/meminfo to get available memory

evetsso:
sysinfo(2) is unaware of cache, but we can still fall back to it in case /proc/meminfo isn't available.

After testing some more on my integrated gfx1035 I realized #595 is too pessimistic.  [sysinfo(2)](https://linux.die.net/man/2/sysinfo) is unable to get the same information that [free(1)](https://linux.die.net/man/1/free) gets, and we're really interested in the "available" number.

So for example, with 32 GiB system memory (16 GiB allocated to GPU) `free` says I have 12 GiB free and 11 GiB cached (23 GiB available).  With the code that only looks at `sysinfo` none of the test cases will run - when we take away the 16 GiB for the GPU from 12 GIB free, there's nothing left.

free(1) ultimately looks at /proc/meminfo to know its numbers so just look there first.

### DIFF
--- a/shared/sys_mem.h
+++ b/shared/sys_mem.h
@@ -122,7 +122,7 @@ private:
             std::string   proc_line;
             while(std::getline(proc_meminfo, proc_line))
             {
-                // meminfo counts in kiB
+                // meminfo counts in KiB
                 if(proc_line.compare(0, 9, "MemTotal:") == 0)
                     total_bytes = std::stoull(proc_line.substr(9)) * 1024;
                 else if(proc_line.compare(0, 13, "MemAvailable:") == 0)
--- a/shared/sys_mem.h
+++ b/shared/sys_mem.h
@@ -111,11 +111,42 @@ private:
         total_bytes = info.ullTotalPhys;
         free_bytes  = info.ullAvailPhys;
 #else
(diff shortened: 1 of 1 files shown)


shared/sys_mem.h
```diff
@@ -111,11 +111,42 @@ struct host_memory
...
+
+        // /proc/meminfo can tell us "available" memory (i.e. free +
+        // buffers + cache)
+        total_bytes = 0;
+        free_bytes  = 0;
+        try
+        {
+            std::ifstream proc_meminfo("/proc/meminfo");
+            std::string   proc_line;
+            while(std::getline(proc_meminfo, proc_line))
```
regan-amd, line 123: Shall we check `proc_meminfo.is_open()` first, maybe?

evetsso: getline will fail if the file couldn't be opened, so I don't think it'll make a difference in behaviour?

af-ayala (approved):

regan-amd (approved):

eng-flavio-teixeira (approved):

Kevonosdiaz (approved):

malcolmroberts (approved):

PR #599, opened 2025-06-14
Fix for possible false positives in testing when using cached results and non-unit scale factor

regan-amd:
Self-explanatory title

Example of current false-positive failure addressed with the suggestion (on gfx1201):
`./rocfft-test --smoketest --seed 4697 --gtest_filter=*4_4_8192*`

Will need to be cherry-picked in hipFFT if/once merged.

### DIFF
--- a/shared/accuracy_test.h
+++ b/shared/accuracy_test.h
@@ -53,6 +53,7 @@ struct last_cpu_fft_cache
     fft_transform_type  transform_type = fft_transform_type_complex_forward;
     bool                run_callbacks  = false;
     fft_precision       precision      = fft_precision_single;
+    double              scale_factor   = 1.0;
 
     // FFTW input/output
     std::vector<hostbuf> cpu_input;
@@ -851,7 +852,8 @@ inline void fft_vs_reference_impl(Tparams& params, bool round_trip)
     std::unique_ptr<StoreCPUDataToCache> store_to_cache;
     if(fftw_compare && last_cpu_fft_data.length == params.length
        && last_cpu_fft_data.transform_type == params.transform_type
-       && last_cpu_fft_data.run_callbacks == params.run_callbacks)
+       && last_cpu_fft_data.run_callbacks == params.run_callbacks
+       && last_cpu_fft_data.scale_factor == params.scale_factor)
     {
         if(last_cpu_fft_data.nbatch >= params.nbatch)
         {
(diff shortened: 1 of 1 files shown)


shared/accuracy_test.h
```diff
@@ -859,6 +859,20 @@ inline void fft_vs_reference_impl(Tparams& params, bool round_trip)
             cpu_input.swap(last_cpu_fft_data.cpu_input);
             cpu_output.swap(last_cpu_fft_data.cpu_output);
             run_fftw = false;
+            if(params.scale_factor != 1.0)
```
evetsso: Another approach would be to add `scale_factor` to `last_cpu_fft_data`, similar to how it already tracks `run_callbacks`.  This would mean the cpu output is still stored scaled in the cache, but at least the logic that manages the cache is aware that this has happened.

IMO that would be simpler than trying to undo the result scaling?

regan-amd: That's what I had first considered but I thought this could trigger more cache flushes than necessary and slow down testing as a consequence, conceptually. That being said, that was assuming randomly-generated values for scale factors while the test base uses only 3 possible scale factor values, as of now: $1.0$ (default), $7.23$ (if set by a parameter generator) or $0.1239$ (set when reading a token that uses `scale`).
So, the likelihood of running into configurations differing only in values of `scale_factor` is significantly reduced compared to what randomly-generated values would produce. Randomizing inputs often exposes less expected errors and could be considered beneficial in the future though, _e.g._, testing for possible losses of accuracy due to (hypothetical) reckless casts.

If we have no intention to change the way the scale factors are set/chosen, I agree it'd be better to make `scale_factor` a member of the cache key. In that case, let me know and I'll go ahead with that option.

evetsso: I don't think we currently have any plans to change scale factors - I deliberately tried to keep it simple enough so that we wouldn't have to try to store the scale factor number in the token, and can't foresee a reason to change that anytime soon.

Another option we could consider is to simply not attempt to cache results if scaling is enabled.  We've only got a handful of scaling tests, so increased runtime isn't likely to be an issue.  And I don't think we'd meaningfully increase coverage by expanding those test cases.

But yeah, adding it to `last_cpu_fft_data` sounds simplest for now.

regan-amd: Thanks, suggestion done in c06207a

evetsso (approved):

af-ayala (approved):

eng-flavio-teixeira (approved):

Kevonosdiaz (approved):

PR #614, opened 2025-06-30
Enable CommAllToAll for Pencil Decompositions using MPI Sub-Communicators

af-ayala:
Summary of proposed changes:

- Currently, when all-to-all communication is not possible (limited to slab decomposition), rocFFT relies on P2P communication. The approach is to gather data into a single GPU, perform the computation, and then scatter the results, as shown in the profile:

image
 
-  The objective of this PR is to add support for All-to-All communication with sub-communicators for Pencil decomposition

### DIFF
--- a/library/src/include/plan.h
+++ b/library/src/include/plan.h
@@ -1,4 +1,4 @@
-// Copyright (C) 2016 - 2023 Advanced Micro Devices, Inc. All rights reserved.
+// Copyright (C) 2016 - 2025 Advanced Micro Devices, Inc. All rights reserved.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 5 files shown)


library/src/include/plan.h
```diff
@@ -134,6 +134,12 @@ struct rocfft_plan_description_t
     rocfft_comm_type comm_type = rocfft_comm_none;
 #ifdef ROCFFT_MPI_ENABLE
     MPI_Comm_wrapper_t mpi_comm;
+    bool               use_subcomm = false;
+    MPI_Comm_wrapper_t subcomm;
+
+    // hard-coded test
```
malcolmroberts, line 140: Why is there a need for a hard-coded test?

malcolmroberts:
From the description, it seems that this PR is designed to do two totally different things: use sub-communicators for pencil, and do something besides gather/scatter for more complicated cases.  If that's the case, this PR should be split in two; otherwise, could you please clarify the intent?

PR #619, opened 2025-07-05
Scope -Xarch_device -O3 to Avoid RTC Cache Errors on Windows

af-ayala:
#solves 1239
Adding -Xarch_device -O3 causes the RTC-compiled kernel binaries to differ, leading to hash or ABI mismatches on Windows. These mismatches currently break kernel reuse from cache on Windows builds during repeated runs. For example:

`rocfft-test.exe --gtest_filter=*rocfft_UnitTest.rtc_cache* --gtest_repeat=100`

This change limits the flag to non-Windows OSes to prevent RTC cache errors during testing.


evetsso:
Can you explain in more detail what's happening here?  A side effect of this change should be that the callback tests will break in a debug build, as mentioned by the comment above your change.

ammallya:

PR #625, opened 2025-07-15
Add partial-pass 3D kernels

eng-flavio-teixeira:
Enables the partial-pass optimization for additional 3D lengths: 

 - {64, 64, 128}
 - {64, 64, 52}
 - {60, 60, 60}
 - {32, 32, 128}
 - {32, 32, 64},
 - {64, 32, 128}

Several fixes to the code generators (StockhamPartialPassKernelCC and StockhamPartialPassKernelRR) were required to support the additional lengths. These fixes include:

 - Changes to data offset calculations.
 - Changes to the LDS-to-register and register-to-LDS generators.
 - Fixes for the partial-pass steps 1-2 and 3-4.
 - Generalization of the local transposition to the non-isometric case.
 - Added thread guards to the LDS and global memory accesses to expand supported wgs and tpt configurations.
 - Improved validation of kernel configuration parameters.
 - Fixes to the partial-pass twiddle generation kernel.

As can be seen in the attached results for MI250X and MI300x, there is a significant performance uplift with the new kernels.

NOTES:

There is still a slight performance regression (~5%) for the previously supported 64x64x64 case. The reason for this regression is due to changes to the local transposition step. Since this calculation now supports the general non-isometric case, the global memory indices transformations are more involved, thus requiring additional operations. The specialization of these calculations to the isometric case will be implemented in a subsequent PR, and this should bring the performance of the newly generated kernels for the 64x64x64 case on par with the old ones.

In addition, the main reason for not enabling the optimization for more 3D-lengths is that the generators currently only support a single partial-pass - in other words, the kernel _factors_pp_ parameter only supports a single radix entry, and this limits considerably the enablement of the optimization due to LDS usage constraints. In another subsequent PR, support for multiple partial-passes will be added, and this should allow us to to enable the optimization for significantly more problems.


eng-flavio-teixeira:
Attached performance results for MI250X and MI300X:

[results_MI250X.zip](https://github.com/user-attachments/files/21422816/results_MI250X.zip)
[results_MI300X.zip](https://github.com/user-attachments/files/21433857/results_MI300X.zip)

ammallya:

PR #626, opened 2025-07-19
Adding auto-allocation logic for rocfft tests

regan-amd:
Following up here upon completion of https://github.com/ROCm/hipFFT/pull/160.

Note:
1. if auto-allocation is used, plans allocate resources at execution for rocFFT whereas thy do it at plan generation for hipFFT;
2. "default" auto-allocation flag for rocFFT tests is equivalent to "off", while "default" auto-allocation behavior for hipFFT tests is equivalent to "on", consistently with the above;
3. If "on" auto-allocation flag is used for a rocFFT test, that would force the underlying plan to allocate resources at execution time. It might be best to have a warning printed by bench in that case?


ammallya:

PR #627, opened 2025-07-22
Enable pencil-to-pencil global transposition with sub-communicators

af-ayala:
Summary of proposed changes:
 
-  The objective of this PR is to add support for All-to-All communication with sub-communicators for uniform Pencil decompositions, as shown in the following figure.
 
image

The proposed updates work as follows:

1. We have added a function that reads (or finds) the input and output grids for `inField` and `outField`. It then returns a transposition plan for pencil decomposition, including brick-to-pencil, pencil-to-pencil, slab-to-pencil, etc. For example: if `in_grid = {4, 8, 4}` and `out_grid = {8, 4, 8}`, it returns the following sequence of intermediate grids: `{4, 8, 4} → {1, 16, 8} → {16, 1, 8} → {16, 8, 1} → {8, 4, 8}` along with a corresponding sequence of transpose types. The selection of intermediate grids is based on a heuristic approach.

2. In the case of uniform pencils; i.e., when there is a uniform volume of messages to send and a subset of neighbors to send to, we create pencil sub-communicators and perform a global transpose using `MPI_Alltoall`.

3. This PR focuses only on pencil-to-pencil transposition. Therefore, we have left the default path intact, which relies on either `MPI_Alltoallv` or `MPI_Alltoall` when the global communicator allows for uniform exchanges (slab case).

4. Implementing an optimized pencil-to-pencil with MPI_Alltoall is not just for performance gains, but also to ensure scalability.

### DIFF
--- a/library/src/plan.cpp
+++ b/library/src/plan.cpp
@@ -1977,7 +1977,8 @@ void rocfft_plan_t::GlobalTranspose(size_t                     elem_size,
                                     std::vector<BufferPtr>&    output,
                                     const std::vector<size_t>& inputAntecedents,
                                     std::vector<size_t>&       outputItems,
-                                    size_t                     transposeNumber)
+                                    size_t                     transposeNumber,
+                                    MPI_Comm_wrapper_t&&       subcomm)
 {
     // All-to-all transpose is preferred as it's faster. This requires
     // that each rank have a single base pointer to send/receive with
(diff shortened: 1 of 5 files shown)


malcolmroberts:
In which cases is there a performance benefit?

library/src/plan.cpp
```diff
@@ -2268,15 +2269,81 @@ void rocfft_plan_t::GlobalTransposeA2A(size_t                     elem_size,
         }
     }
```
malcolmroberts, line 2507: This is mpi-specific; why shouldn't it be in rocfft_mpi.h?

library/src/tree_node.cpp
```diff
@@ -904,6 +904,85 @@ void CommGather::Print(rocfft_ostream& os, const int indent) const
     }
 }
```
malcolmroberts: MPI-specific code is better in rocfft_mpi.h

library/src/tree_node.cpp
```diff
@@ -904,6 +904,85 @@ void CommGather::Print(rocfft_ostream& os, const int indent) const
...
+    case 1: // Transposing along Y → group by XZ
+        return x * grid[2] + z;
+    case 2: // Transposing along Z → group by XY
+        return x * grid[1] + y;
+    default:
+        throw std::runtime_error("Invalid split_dim");
+    }
+}
+
+int CommAllToAll::calculate_subcomm_size(const std::array<int, 3>& grid, int split_dim)
```
malcolmroberts: Only for 3D transforms?

malcolmroberts (changes requested):
I'd really like to avoid macros, is possible.

Of bigger concern is that use_intermediate_slabs seems to only be false.

malcolmroberts:
Feature cutoff deadline is today; we can discuss whether we want to move this to 7.2.

af-ayala:
In terms of performance, gains are observed starting at 2 nodes, with 16 GPUs, we can observe ~9% gains for symmetric sizes such as $512^3$.  For 32 and 64 GPUs, gains around 13%.

ammallya:

